PR #52, opened 2020-10-01
Update index.js

stdye:
Trying to shift spectrum plot to top of tab...


stdye:
Changed order of export arguments in index.js. This introduced no errors but did not change the order of the cards. Hmmm... order must be dictated elsewhere.

PR #53, opened 2020-10-01
Update physics-plots.jsx

stdye:


DocOtak:
Because of the "special stuff" going on with one of the titles, I had to tell the webapp that these are "string literals" and should not be interpreted in any special way. So it needs a little extra care when trying to add something like line breaks.

Issue #48, opened 2020-09-30
Reactors tab: add card for listing Fission Fractions/Energies


stdye:
My attempt at the html code for the table with guesses for importing variables- please comment

Note: It shouldn't be necessary to specify the fission fraction values as I tried below. They should be set elsewhere.

```jsx
import React from "react";
import { Card, Table } from "react-bootstrap";

import const {Fission_Fractions: { [type: string]: FissionFractions } = {
   LEU: {
     U235: 0.56,
     U238: 0.08,
     PU239: 0.3,
     PU241: 0.06,
   },
   GCR: {
      U235: 0.7248,
      U238: 0.0423,
      PU239: 0.2127,
      PU241: 0.0202,
   },
   PHWR: {
      U235: 0.52,
      U238: 0.05,
      PU239: 0.42,
      PU241: 0.01,
   },
   LEU_MOX: {
      U235: 0.39,
      U238: 008,
      PU239: 0.42,
      PU241: 0.11,
   },
} from "../physics/fission-fraction";

export const FissionFractionPane = () => {
  const tableProps = { style: { width: "auto" }, borderless: true, size: "sm" };

  return (
    <Card>
      <Card.Header>Fission Fractions and Energies</Card.Header>
      <Card.Body>
          <p>Mid-cycle fission fractions
          </p>
          <p>
          GCR: R. Mills, private communication, Dec. 12, 2018.
          PHWR: M. Chen, private communication, Aug. 14, 2017.
          PWR, BWR; MOX (PWR, BWR): M. Baldoncini et al., "Reference worldwide model for antineutrinos from reactors," Phys.  Rev. D91, 065002 (2015).
          </p>
          <p> Fission energies
          </p>
          <p>
          V. I. Kopeikin et al., "Reactor as a Source of Antineutrinos: Thermal Fission Energy," Phys. Atom. Nucl. 67, 1892 (2004).
          </p>          
        <Table {...tableProps}>
          <thead>
            <tr>
              <th></th>
              <th><sup>235</sup>U</th>
              <th><sup>238</sup>U</th>
              <th><sup>239</sup>Pu</th>
              <th><sup>241</sup>Pu</th>
            </tr>
          </thead>
          <tbody>
            <tr>
              <td>
              Q (MeV)
              </td>
              <td>{qf235}</td>
              <td>{qf238}</td>
              <td>{qf239}</td>
              <td>{qf241}</td>
            </tr>
            <tr>
              <td>
              f (GCR)
[...]

PR #66, opened 2020-10-02
Add start of reactor fission fraction table

DocOtak:
See #48


stdye:
Looks good! Just need to add PWR, MOX, and PHWR fission fractions, then the content is set.

PR #67, opened 2020-10-02
Update reactors-fission.jsx

stdye:


src/ui/reactors-fission.jsx
```diff
@@ -52,9 +56,18 @@ export const FissionFractionPane = () => {
                 <td>{FISSION_FRACTIONS.GCR.PU239}</td>
                 <td>{FISSION_FRACTIONS.GCR.PU241}</td>
               </tr>
+              <tr>
+                <td>
+                f (GCR)
```
DocOtak, line 61: Probably want this to be `f (PHWR)`

PR #69, opened 2020-10-02
Update constants.ts

stdye:


DocOtak (approved):
Looks good to me, any idea where the original numbers came from?

PR #70, opened 2020-10-02
replicate table V

DocOtak:


stdye:
Good thing we checked fission fractions and fission energies! We need to be careful not to mix up power factions and fission fractions. See Eq.s 19 and 20 in Calculations report and let me know if the web app calculation is doing what we want here.

stdye:
I think I have not been careful enough with the distinction between power fractions and fission fractions. The PWR, BWR fractions (0.560,0.080,0.300,0.060) from Baldoncini et al. are power fractions.

Power fractions plug directly into Eq. 18 to get the total rate.

The fractions I got from Mark Chen (PHWR) and from Robert Mills (GCR) are, I think and will recheck the emails, are fission fractions. If so, these need to be converted to power fractions, using Eq. 20, before plugging into Eq. 18.

Meanwhile, I will also try to make sense of the MOX values.

Here is the Baldoncini et al. paper-

[Baldoncini_atal_2015_PRD.pdf](https://github.com/geoneutrinos/reactors/files/5320543/Baldoncini_atal_2015_PRD.pdf)

DocOtak:
I'll be looking into this one this afternoon. We can either merge this in to get the table displayed, or I can update it once we get the fission vs power fraction issues sorted.

stdye:
The issue of power fractions vs fission fractions is resolved. See the paragraphs on page 5 with equations 19 and 20 embedded and ending with Table V.

[Web_app_calculation.pdf](https://github.com/geoneutrinos/reactors/files/5329475/Web_app_calculation.pdf)

Because the fission energies are not too different in value, the values of fission fractions and those of the corresponding power fractions (Eq. 20) are not too different.

Table V lists power fractions for PWR/BWR, pure MOX, and PWR/MOX at 70/30, and also lists fission fractions, as given to us by Robert Mills and Mark Chen, for GCR and PHWR.

Following our decision to only use input data from original sources, the fission fractions need to be internally converted to power fractions using Eq. 20 prior to executing Eq. 19. This probably takes a few more lines of code to implement but it sounds rather simple to me.

As it stands now the table that is live on the site is correct but incomplete. Launching #70 publishes incorrect information in a complete table. My preference is to update #70 and then merge.

stdye:
We need to:

1) add "Mid-cycle" before "fission energies"
2) change the title to be "Fission Energies and Power/Fission Fractions"
3) start each reference on a new line

on this new card.

stdye (commented):
See the comments I posted to the pull request

DocOtak:
OK, I've confirmed that the web app has a slightly rearanged (for speed/caching reasons) implimentation of eq 17, 18, and 19. Internally it is looking at a constant called "FISSION_FRACTIONS". With the exception of GCR and PHWR types, these are actually power fractions, which is what the functions want.

There is some renaming and conversions to do...

stdye:
Symbols p and f reversed?

PR #76, opened 2020-10-06
Impliment cost theta plot

DocOtak:
So here is the first crack at it, naturally it needs refinements. Is it what is expected?

Since our data source has multiple cores at the same site tracked independently, often the largest signal is going to be from other cores at the same site (cos = 1)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -14,6 +14,7 @@ import {
   StatsPanel,
   DetectorLocationPane,
   //Reactors Tab
+  CoreDirectionSignalPlots,
   FissionIsotopeSpectraPlots,
   CoreIAEARange,
   CoreList,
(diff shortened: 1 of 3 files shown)


stdye:
It would be helpful to use marker colors matching the colors for the core types, like we have done on the map and the polar plot. In particular, I am trying to make sense of the cos#theta plot for SNOLAB.

DocOtak:
Yeah I think I can make the colors happen.

Re making sense of the plot, the closest reactor complex is BRUCE which has something like 8 cores with different signals. It helps if you switch the plot to "show closest data on hover" (third icon from the right).

stdye:
Try setting location to IMB. Perry core does not show on plot

DocOtak:
Did a bunch of label cleanups, should be consistent now, though might need some tweaks in sigfigs.

For the PERRY core, what is the angle between the closest core and itself? PERRY is a single core reactor site....

stdye:
Perry is the closest core at IMB

PR #83, opened 2020-10-08
Update physics-constants.tsx

stdye:
Here is a start on the content of the Physical Constants table.

I am not sure what name you want for the table so I left it as "table????" 

I am not sure how you want to pass the values of these constants so I just entered the values that are in the web app calculations report

Of course, all checks will fail but it gets things moving albeit with perhaps some squandering of build time.

### DIFF
--- a/src/ui/physics-constants.tsx
+++ b/src/ui/physics-constants.tsx
@@ -1,16 +1,18 @@
 import React from 'react';
 
-import { Card } from 'react-bootstrap';
+import {FERMI_COUPLING_CONSTANT, HBAR_C, ELECTRON_REST_MASS, WEAK_MIXING_ANGLE} from '../physics/constants'
+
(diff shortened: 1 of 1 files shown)


DocOtak:
We might need to update the weak mixing angle value. We app is currently using 0.2313 (maybe from [arXiv:hep-ph/0212134](https://arxiv.org/abs/hep-ph/0212134)) and not 0.23153.

stdye:
Here is the reference link for the description section of the card-

https://pdg.lbl.gov/2020/reviews/rpp2020-rev-phys-constants.pdf

See double dagger note at bottom

stdye:
Can you quickly clean up my mess with the symbols in the table header? If not, I can work on it later.

stdye:
I am not sure how to make changes without first merging the pull request and then editing the new merged file. Maybe it doesn't matter but what about the build time?

PR #91, opened 2020-10-09
Update physics-osc-params.tsx

stdye:
Tried to rotate table to save space and to update the reference for the new values. Must have buggered something.


stdye:
Should be easy to fix and update with the new values.

Also need to add entries in the table for <P_e> that we use for geo-neutrinos. These can go in the last column.

stdye:
Note that PDG provide delta m^2_32. This means that our listed value of delta m^2_31 is derived from input data.

Note that for Inverted Mass Ordering delta m^2_32 and delta m^2_31 are NEGATIVE and should be listed as such.

PR #92, opened 2020-10-12
Update osc params from PDG

DocOtak:


stdye:
Almost there. PDG changed from giving dmsq31 to now giving dmsq32. This changes the primitive value from dmsq31 to dmsq32. For example, the new table lists dmsq32(inv) = -2.5089 but given as -2.509 by PDG

PR #95, opened 2020-10-21
Implement angular differential cross section plot

DocOtak:
Wasn't actually sure what to call this plot...

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -29,6 +29,7 @@ import {
   PhysicsConstants,
   CrossSectionPlots,
   DifferentialCrossSectionPlots,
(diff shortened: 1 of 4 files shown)


stdye:
Tracking down factor 2 discrepancy between Fig 3 and your plot

stdye:
Fig 3 nuebar peaks at cos\theta ~ 0.95 are factor of 2 greater than those on web app

DocOtak:
I tried implementing this by actually multiplying Eq 11 and 13 in the code, eq 6 is used to get the Te input for eq 11...

PR #97, opened 2020-10-22
Add T_min param to crossSectionElasticScattering function

DocOtak:
Right now this is a work in progress PR, it hard codes a T<sub>min</sub> of 5 MeV for ES interactions.


stdye:
With T_min = 2.5 MeV solar tab lists R_sol = 295.56 NIU while my calculation gives 279.093 NIU. Previously, results for T_min = 5.0 MeV and T_min = 0 agreed.

stdye:
Due to the sharp decrease in the ES interaction rates when applying a cut on T_min, it becomes evident that more decimal places are needed on the reported rates on the spec stats pane (e.g. R_geo shows as 0.0 NIU even though a blip in the spectrum is apparent). It would be very nice if the rate values could always be shown with 3 and only 3 significant figures (i.e. 127 NIU, 32.6 NIU, 8.32 NIU, 0.495 NIU, 0.0642 NIU, etc.)

DocOtak:
It is possible using [`toPrecision`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision) if you are happy with "large" numbers appearing in exponential form:

```console
> (1359.0).toPrecision(3)
< "1.36e+3"
> (1359.0).toPrecision(4)
< "1359"
```

stdye:
toPrecision(4) is good

stdye:
It looks like this is working. It will be great to have the UI working.

Here is a plot of my calculated R_sol for T>T_min in NIUs.
 
[8B_rate_temin_NIU.pdf](https://github.com/geoneutrinos/reactors/files/5498242/8B_rate_temin_NIU.pdf)

DocOtak:
There are, as far as I can tell, two more places where things depend directly on the cross section functions:
1. Solar neutrinos
2. the rateToFlux calculations (used with the Sanshiro geonu data)

stdye:
Sounds right

stdye:
Solar neutrinos are a serious background to the ES signal from reactors. Geo-neutrinos can be ignored in most scenarios.

stdye:
Although it is not possible to set the T_min slider exactly where I want, I can get pretty close. So the good news is: The nuebar and nuxbar rates for the sum of the two Hartlepool cores using 2018 average annual LF as given by the web app "agree" with my calculations at T_min = 0, 2, 3.5, 5 MeV. I had to do a scroll over on the cos(theta) plot to get precision on the rates better than on the spectrum stats pane (see previous comments on this thread). Considering not being able to set T_min on the values in my calculations and the somewhat coarse precision on the rates reported on the cos(theta) plot, our rates "agree."

stdye:
????  NIU rates for ES 31.0 + 3.4 = 42.1

Set Detector to IMB
Set LFs to 2018
Turn off all cores except Perry-1
Turn off Geonus
Set T_min to 2.0 MeV
Set xsec to ES antinu- rate = 42.1 NIU (disagrees with my Fortran calculation)
Set xsec to ES nuebar- rate = 31.0 NIU ("agrees" with my Fortran calculation)
Set xsec to ES nuxbar- rate = 3.4 NIU ("agrees" with my Fortran calculation)

DocOtak:
Good catch re ES total not matching the sum of the respective parts.

stdye:
I have trouble sliding in a precise value of T_min. Can we have a control that offers better accuracy? One like the controls in the Mantle Flux Inputs would be good with default value set to 0.00 MeV and increments every 0.01 MeV (10 keV). Users can also just type in the value they want rather than struggling with the slider and still not getting the value they want. Thank you

stdye:
[Web_app_calculation.pdf](https://github.com/geoneutrinos/reactors/files/5713469/Web_app_calculation.pdf)

see Table XV for solar rates at different T_min values
see Figure 22 for spectra at different T_min values

stdye:
With the more correct sin^2(theta_W) value the ^8B solar nu rates at T_min = 0, 2, 3.5, 5 MeV are

464.9, 318.4, 221.0, 139.8 NIU

[Web_app_calculation.pdf](https://github.com/geoneutrinos/reactors/files/5713636/Web_app_calculation.pdf)

stdye:
What does the green check mark at the far right of the T_min box signify?

It went away after I typed in "10" and then the plots kind of went haywire.

DocOtak:
It's trying to let you know that your input is valid... but doesn't seem to update itself the way it should (e.g. if you type a letter in a browser that will let you, it _should_ turn red).

stdye:
Aha, that would be useful when it works as designed.

stdye:
Why not apply the T_min cut to IBD, too? It should be pretty easy to add to the rates/spectra (I think it is in the Calculator already).

T_e+ (IBD) = E_nuebar - E_threshold

stdye:
Slider works well. BUT weirdness happens after setting to 10 MeV. Maybe try setting maximum to 9 MeV

stdye:
Can we get sliders on the mantle flux values, too?

Flux- default 1e6, max 5e6, min 1e5
Th/U ratio- default 3.9, max 10, min 1
K/U ratio- default 1e4, max 1e5, min 1e3

stdye:
On Safari browser any T_min setting over 7.0 MeV makes the green fill spill out from under the plot toward the origin.

DocOtak:
The plot thing with T_min >7MeV is some sort of rendering bug in plotly. I tried removing all the 0 data points being plotted and that green bar still showed up. Turning the fill off did seem to work.

Rendering issues aside, is it ok scientifically to limit the T_min to a max of 7MeV?

Let's address the IBD T_min and Sliders for geonu in another pull request so we can land this one soon (today?)

stdye:
The sensible maximum T_min depends on the endpoint energy of the neutrino signal and the sensitivity of the detector.
A maximum T_min of 2.5 MeV is okay for geo-neutrinos- This cuts out K and Th but not U
A maximum T_min of 7 MeV is okay for reactor antineutrinos- This cuts out geo-nus and most of the reactor antineutrinos
A maximum T_min of 10 MeV is okay for 8B solar neutrinos- This cuts out geo-nus and reactor antineutrinos

stdye:
We can live temporarily with a maximum T_min of 7 MeV to claim some progress and retire this pull request.

The associated issues on this thread can be transferred to New Issues.

PR #103, opened 2020-10-31
Update physics-constants.tsx

stdye:


stdye:
The idea is to add "... for ES" to the card title. A new "Physical Constants for IBD" card would then be added above that lists m_n, m_p, and m_e. Alternatively, one could add a second table above the existing one on the existing Physical Constants card for listing m_n, m_p, and m_e.

stdye:
This is now obsolete with the addition of the new table under the Input Tab- Table: Particle Masses

Issue #12, opened 2020-05-10
Increase model bin resolution


DocOtak:
Some input data (e.g. the Sanshiro geoantineutrino spectrum data) is available in a higher resolution than the 10KeV bins we have been using. It should be possible to use this higher precision in some places.

Things to investigate:
* Where in the code we assume 10 KeV bin width
  * this would be things that have numbers like 0.005 or i/100
* ways to do a "high precision run" for things like downloading the data (i.e. swap bin size on the fly)
  * maybe only use 10KeV bins when in certain modes like follow cursor

PR #135, opened 2021-01-06
Add boron8 to download output when ES interaction selected

### LINKED ISSUE #129: Add solar nu 8B reaction rate spectrum to output tab

DocOtak:
Closes #129

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -48,15 +48,15 @@ import { presets, detectorENUProjector } from "./detectors";
 import { getCrustFlux } from "./crust-model";
 import { mantleGeoSpectrum } from "./mantle";
 
-import {crossSection as initialCrossSection} from "./physics/neutrino-cross-section"
-import { crossSectionReducer} from "./physics/neutrino-cross-section"
+import { crossSection as initialCrossSection } from "./physics/neutrino-cross-section";
+import { crossSectionReducer } from "./physics/neutrino-cross-section";
 
-import {oscillation as initalOscillation} from "./physics/neutrino-oscillation";
-import {oscillationReducer} from "./physics/neutrino-oscillation";
+import { oscillation as initalOscillation } from "./physics/neutrino-oscillation";
+import { oscillationReducer } from "./physics/neutrino-oscillation";
 
-import {defaultBoron8} from "./solar"
+import { defaultBoron8 } from "./solar";
 
-import {PhysicsContext} from "./state";
+import { PhysicsContext } from "./state";
 
 import "leaflet/dist/leaflet.css";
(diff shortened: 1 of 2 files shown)


DocOtak:
It would be nice to add boron8 data in 10KeV bins.

stdye:
Major issues:

1) Solar neutrino 8B (sol8B) reaction rates extend to about 16 MeV while the big csv download ends at 10 MeV. The rates at neutrino energy > 10 MeV need to be in the download. A 15 MeV solar neutrino that just glances an atomic electron can give an electron just a few MeV of kinetic energy or less.
2) The sol8B rates in the trial csv file are many orders of magnitude too small if the rates are NIU

Minor issues:

1) The label for the header column on the solar nu rates could mirror the geo-neutrino labels if it is "sol8B." As is, the source of the neutrinos is not fully specified
2) Add units to the labels on the column headers of the csv file. Except on the energy column (MeV), the units of the values in the columns are missing in the labels on the column headers

stdye:
Re: It would be nice to add boron8 data in 10KeV bins

There are several mismatches in the output data for the reaction rates from the different neutrino sources.

1) Solar 8B data are in 100 keV bins and extend to almost 16 MeV. The output file carries many repeated values.
2) Geo-neutrino data cut off at various levels below 4.4 MeV (40K ~1.3 MeV; 232Th ~ 2.3 MeV; 238U ~ 4.4 MeV). The output file carries many 0s.

A related topic is Issue #12 to increase the model's bin resolution from 10 keV to 1 keV.

One output file with all the data is getting bothersome.

What are your thoughts on making the output data available for downloading in 3 separate files?

1) The solar 8B file could have 100 keV bins from 0 to ~16 MeV.
2) The geo-nu files could retain their 10 keV bins but range in energy from 0 to ~4.5 MeV
3) The reactor antineutrino files could retain the 0 to 10 MeV energy range but increase resolution to 1 keV bins

DocOtak:
Separate files are the way to go unless we match the binning of the rest of the model.

stdye:
The 8B solar neutrino spectrum from Winter et al comes in 100 keV bins

The geo-neutrino spectra from Enomoto come in 1 keV bins

The initial implementation of the reactor spectra employs 10 keV bins

Oy vey!

stdye:
This looks good! The separate file works well. I recommend adding "8Bsolar" or a similar identifier to the filename.

How about separating out the geo-nu output, too?

DocOtak:
I thought I added it to the filename... but seems it was just the download button itself.

stdye:
Let's make a separate download file for geo-nu. I will generate a new issue

PR #137, opened 2021-01-08
Update solar-plots.jsx

stdye:
We changed quantity in plot without changing plot title


stdye:
Eventually, we will add cards for 7Be, CNO, hep, etc. neutrinos

PR #140, opened 2021-01-14
Update App.js

stdye:
It took me a while to find the text under the "About" tab. This attempts a revision now that the new version of the paper is on arXiv.


stdye:
It looks like this worked. We should include a simple acknowledgment of support by Livermore.

"This work was supported in part by Lawrence Livermore National Security, LLC."

stdye:
Ugh... I should be able to edit my edit before adding it to the online version.

PR #142, opened 2021-01-15
Update index.ts

stdye:
Try adding fission fractions for VVER core


stdye:
This seems to have worked. I can now try adding fission fractions for beginning and end of operation cycle

Issue #133, opened 2020-12-22
Antineutrino signal evolution with fuel Burn-up


stdye:
This Issue addresses the deliverable: **the inclusion of evolving reactor fission fractions with burn-up**.

A description of how to begin work on this issue is in Section III, including Table VIII and Figure 11, in this version of the web app calculation report-

[Web_app_calculation.pdf](https://github.com/geoneutrinos/reactors/files/5731757/Web_app_calculation.pdf)

stdye:
This should be an easy issue to retire. If you have a plan in mind, please share it. If not, consider the following:

1) Add new sets of fission fractions (I am guessing to reactors/src/reactor-cores/index.ts) so they can be linked to the "Fission Fraction Suggestions" pull down menu for the "Custom Core"

2) The new sets of fission fractions, which are given with references in Table VIII, are for AGR and VVER cores and include values for the start, middle, and end of the operating (fuel) cycle

To study burn-up, users can then replace an IAEA core, which they turn off, with a custom core added at the IAEA core's Lat,Lon,Elev with the IAEA core's power; download the spectrum, put in new fission fractions, download the new spectrum, etc.

PR #143, opened 2021-01-15
Update index.ts

stdye:
Now add all fission fractions from Table VIII in the document posted on arXiv- see About tab


stdye:
Seems like things are still working. Give it a shot... This links to Issue #133

PR #144, opened 2021-01-15
Update geonu.jsx

stdye:
Attempt to prevent negative flux values


stdye:
Adding min="10000" seems to work but this positive value does not allow turning off the mantle flux altogether. Will try min="0" for the mantle flux and the Th/U and K/U ratios, which also currently allow non-physical negative values

PR #146, opened 2021-01-15
Update geonu.jsx

stdye:
Attempt to change mantle 238U flux input values to slider


stdye:
Oops! Something not right. Kicking this upstairs ;)

PR #147, opened 2021-01-15
Update reactors-plots.jsx

stdye:
Attempt to change y-axis on Core Direction Plot to a log scale. The objective is to prevent a sedimentary layer of cores along the x-axis as is present using a linear scale on the y-axis.

The /physics-plots.jsx was used as an example. Fingers crossed ;)


stdye:
It looks like this worked. The "autorange: True" argument seems to handle values equal to 0. 

An improvement would be using "toFixed(4)" rather than "toFixed(3)" for the rates displayed in the hover-over text box. The values at the bottom of the plot come up as 0.000 even at different levels.

PR #149, opened 2021-01-15
Update reactors-plots.jsx

stdye:
See if the value of the signal for each core displayed on hover can be shown as an exponential- toExponential(2). Given that signals differ by many orders of magnitude, the toFixed() is not practical. With uncertainty in the signal conservatively estimated at ~5%, 3 significant figures is appropriate.


stdye:
No problems that I can see. Take it live and hope for the best, I guess

PR #150, opened 2021-01-16
Update output-download.jsx

stdye:
Try adding a dummy button to the Output.


stdye:
That didn't seem to break anything. Next step...

PR #151, opened 2021-01-16
Update output-download.jsx

stdye:


stdye:
Now just have to strip the GeoNu output from the main file and package it for downloading with the new button

PR #152, opened 2021-01-16
Update output-download.jsx

stdye:
Attempting to strip out GeoNu data and build separate output file for download.

Made filename prefix "Enu" more descriptive:
Enu -> AntiNu for Reactor/Geo
Enu -> SolarNu for Solar
Enu -> GeoNu for Geo (New, I hope)

It will be a miracle if this works. What the heck, I gotta try.

PR #153, opened 2021-01-16
Update output-download.jsx

stdye:
Now try removing GeoNu data from "downloadData" so it only has Reactor data and remove "/Geo" from the buttonTitle


stdye:
Wow, the editing the edited file takes way less time to determine if the code compiles

Apparently, none of the checks were needed.

stdye:
How do I check though? There is no "Details" button to take me to the trial version.

DocOtak:
This is a pull request which changes the other pull request... so it's, probably confusing. When editing an existing PR, you want to commit directly to that "branch" and not open another PR.

stdye:
Ah. Can you check if it is okay?

DocOtak:
Closing this in favor of working on it in #152 (which should be updated/building)

PR #154, opened 2021-01-16
Update plot.jsx

stdye:
Try to make the angular axis (azimuth) use the solar convention of CW from North like in the paper on arXiv.

### DIFF
--- a/src/ui/plot.jsx
+++ b/src/ui/plot.jsx
@@ -172,7 +172,7 @@ export function CoreDirectionPlot({ cores, detector }) {
         dtick: 45,
         tickmode: "array",
         tickvals: [0, 45, 90, 135, 180, 225, 270, 315],
-        ticktext: ["90 &deg;", "45", "0", "315", "270", "225", "180", "135"],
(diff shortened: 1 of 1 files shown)


stdye:
Oops... hoping to avoid reading the whole manual, I copied what I found at this site-

https://plotly.com/javascript/polar-chart/

in the section Polar Chart Directions

stdye:
Hmmm... maybe "angularaxis" is not recognized because "scatterplot" has not yet been declared. If this is the reason, then the problem could be fixed by distributing the "angularaxis" commands to each of the data declarations

DocOtak:
What a fantastically unhelpful syntax error. The actual problem was a single quote `'` instead of a double quote `"` to close the string. This as the effect of causing the string to not actually be closed. I'm sure there is some logical reason it gave the error as expecting a comma somewhere, but you'd need to dig into the JS spec for that.

DocOtak:
Before it's merged... do you want degree symbols on the tick labels?

stdye:
Of course, degree symbols would be great! And the footnote about azimuth at the bottom right of the plot needs updating- N at 0 deg, E at 90 deg, etc.

stdye:
Plot looks good now. Hopefully, there is no problem with this being stale. It shows merging into an old version. Several changes since this was started.

PR #155, opened 2021-01-16
Update reactors-corelist.jsx

stdye:
Some minor formatting of data on core list


stdye:
toFixed() changes make no sense for distances > 100 m due to the precision inherent in the calculation

PR #160, opened 2021-01-18
Update neutrino-cross-section.ts

stdye:


stdye:
Well, I finally broke the site!

The names I changed might be referenced elsewhere in the code and now they do not match up?

stdye:
What is wrong here, sensei?

DocOtak:
Hm... looks like the "CrossSectionPlots" was relying on the exact name, I suspect this was a hold over from when we didn't have a way to deal with the total cross sections.

DocOtak:
Be aware that we currently have a "special case" name of `"Elastic Scattering: Neutrino"` for the solar neutrinos that is not user facing. The `XSNames` object affects the UI on the site.

stdye:
Very good!

Issue #148, opened 2021-01-15
Reactor Cores data cleanup


stdye:
This gathers several issues with the data associated to various reactor cores. It closely relates to #134, which asks for written instructions for the annual reactor core update.

**Change type for 3 US cores #123**

SUSQUEHANNA-1 from PWR to BWR
THREE MILE ISLAND-1 from BWR to PWR
SUMMER-1 from BWR to PWR

**Sanmen-2 latitude incorrect #122**

Move Sanmen-2 to same latitude (29.1011 N) as Sanmen-1. The current location is in the sea.

**Tianwan-1,2 and Tianwan-3,4 have very different and questionably high elevations #101**

Tianwan-1,2 have elevations of 238.3 m and Tianwan-3,4 have elevations of 91.3 m. However, all four cores appear to be located at the same complex on the coast of the Yellow Sea.

Notes: 

1) Core types (#123) and latitudes (#122) appear to be taken from the INFN-provided databases. I cannot edit and save Excel files. MS wants an annual fee so I no longer subscribe to MS Office applications, such as Excel. I think I can open the file with Numbers, edit, and export back to Excel. It is not clear how to edit "DB2019.xls". It looks like one needs to download, edit, and replace.

2) Where do we get core elevations (#101)? From UltraLytics? It is not clear which data file to edit.

stdye:
screen shot showing Sanmen-2 misplaced a full degree north of Sanmen-1

[Sanmen_cores_latitude.pdf](https://github.com/geoneutrinos/reactors/files/7003573/Sanmen_cores_latitude.pdf)

and a zoomed in screen shot of the Sanmen-1 site showing Sanmen-1 and Sanmen-2 labels on the Leaflet map

[Sanmen-1_close_up.pdf](https://github.com/geoneutrinos/reactors/files/7003585/Sanmen-1_close_up.pdf)

How can we relocate this misplaced core?

DocOtak:
I think it's better to have a list of "corrections" that get applied, I've implemented this over in #253

PR #167, opened 2021-01-18
DB2019

stdye:
My first try at uploading a file to the repository.

This file has the corrected core types for the Susquehanna, Three Mile Island, and Summer cores and the corrected latitude for the Sanmen core as described in Issue #148.

The lat and lon formats are right justified rather than center justified.


stdye:
Not sure how to proceed from here. A ReadMe file would be helpful here. There are very few comments anywhere in the code.

PR #168, opened 2021-01-19
Update reactors-corelist.jsx

stdye:
As a diagnostic for figuring out why the core signals and spectrums are different between the web and my Fortran code.

Trying to display core distances and LF with more decimal places. Distance is evident on toFixed() but LF is not. Gonna guess and try setting p=2 in <Num v={lf} p={2} func={(v) => v * 100}


stdye:
Lucky guess! Any issues with allowing this minor change?

PR #169, opened 2021-01-19
Add experimental "raw" data value viewer

DocOtak:
This change allows you to double click on numeric value that use the `<Num />` component to see the underlying data value at full precision.

### DIFF
--- a/src/ui/index.js
+++ b/src/ui/index.js
@@ -39,8 +39,9 @@ export const Visible = ({ children }) => {
 }
 
 export const Num = ({ v, p, func }) => {
+  const [raw, setRaw] = useState(false)
   if (func === undefined) {
     func = (v) => v;
   }
-  return <span title={v.toString()}>{func(v).toFixed(p)}</span>;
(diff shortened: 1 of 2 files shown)


stdye:
Wow, that's cool!

My current hunch is that we are using different neutrino energy values to calculate oscillations. I am using the middle of energy bin. It looks to me like the web might be using the edge of energy bin. Can you please check?

DocOtak:
Reactor core calculations appear to be using the bin center values: https://github.com/geoneutrinos/reactors/blob/c7bddb780c3af10d48c22c7138a46a4ae55236c3/src/reactor-cores/index.ts#L26

DocOtak:
Should we expand on this PR? Maybe combine with #173... if not we can just close.

stdye:
This feature is really useful for debugging. Distance needs to be displayed with precision no greater than +/- 1 m. Most distances are only calculated (stored?) to the nearest km.

DocOtak:
This change ended up sneaking in another way.

PR #171, opened 2021-01-19
Fix not using the bin center for oscillation functions

DocOtak:


stdye:
I used 2018 LFs for my signal comparison. I expect this brings things into better alignment.

DocOtak:
Does it get rid of the wiggles?

stdye:
[signal_ratio.pdf](https://github.com/geoneutrinos/reactors/files/5839815/signal_ratio.pdf)

Not completely but they are definitely smaller now.

stdye:
Not sure what's up with the Boulby curve at high neutrino energy. Your thoughts?

PR #174, opened 2021-01-28
Add preliminary signal ratio thing

DocOtak:
This adds a super preliminary experimental visualizer for the ratio between a core signal at two locations. The plot isn't labeled and the are no instructions yet (since maybe it isn't actually useful) so try the following:

1. When the web app first loads (default to Boulby), go to the "experimental" tab and click the snapshot button.
2. Zoom way into where Boulby mine is
3. Right (or control on a mac) click on the map and select "Follow Cursor"
4. The plot will show the ratio between the snapshot and the current location signal

For Morton:
1. Select Morton from the presets, maybe zoom way into that location
2. Click the snapshot button.
3. Then same steps 3 and 4 above


stdye:
Awesome!!!

stdye:
Crashes site when "Turn off all cores" is selected

In addition to the ratio of the signals, users may want other indicators such as percent change- difference divided by average times 100- (A-B)/(A+B)/.02

stdye:
Neutrino oscillations wiggle according to sin^2(kx), where k (=1.27*delm2/E_nu) is the wave number and x (=L) is the distance. Thus plotting the differential rate (dR/dE) vs 1/E_nu rather than E_nu should reveal equally spaced wiggles. Of course, neutrino oscillations involve several wave numbers- delm2_21, delm2_32, delm2_31- producing a rich spectrum. 

With two detectors the user is exploring dual path interferometry. However, due to sin^2 there are no negative values and no destructive interference (adding signals is probably not very useful). Plotting the ratio is good. Adding an option for plotting the fractional change (or percent change in previous comment) could also be helpful.

PR #176, opened 2021-02-06
Display Reactor LF date interval on detector stats pane

### LINKED ISSUE #175: Add LF info for the signal from IAEA cores on spectrum stats pane
reactors/detector-stats.jsx

line 111 <h6>IAEA Cores</h6>

Add Year-Month range for the load factor data used to calculate the signal.

This would display similar to the way we show the reaction type and cross section info on the Spectrum Stats card

DocOtak:
Closes #175

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -192,7 +192,7 @@ function App(props) {
             <Tabs unmountOnExit={false} defaultActiveKey="detector">
               <Tab eventKey="detector" title="Detector">
                 <Visible>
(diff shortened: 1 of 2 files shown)


stdye:
Tune up- Look what happens when "Use 100% load for all cores" or "Turn off all cores" is selected.

stdye:
Sorry- I forgot to not merge before all tune-ups finished

PR #177, opened 2021-02-06
Initial progress on Analemma plot

DocOtak:
Progress so you can see it.

Check out the solar tab in this PR's preview.

I still need to get the coordinates correct (detector and this solar library I'm using doesn't have the same origin and and some other stuff).

Hold off on hitting that merge button while things still aren't correct.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2760,6 +2760,11 @@
       "resolved": "https://registry.npmjs.org/@types/stack-utils/-/stack-utils-2.0.0.tgz",
       "integrity": "sha512-RJJrrySY7A8havqpGObOB4W92QXKJo63/jFLLgpvOtsGUqbQZ9Sbgl35KMm1DjC6j7AvmmU2bIno+3IyEaemaw=="
     },
+    "@types/suncalc": {
+      "version": "1.8.0",
+      "resolved": "https://registry.npmjs.org/@types/suncalc/-/suncalc-1.8.0.tgz",
+      "integrity": "sha512-1Bx7KgoCLP8LuKaY9whWiX0Y8JMEB9gmZHNJigainwFuv3gEkZvTx0AGNvnA5nSu1daQcJDKScm9tNpW/ZjpjA=="
+    },
     "@types/tapable": {
       "version": "1.0.6",
       "resolved": "https://registry.npmjs.org/@types/tapable/-/tapable-1.0.6.tgz",
@@ -18143,6 +18148,11 @@
           }
         }
       }
+    },
+    "suncalc": {
+      "version": "1.8.0",
+      "resolved": "https://registry.npmjs.org/suncalc/-/suncalc-1.8.0.tgz",
+      "integrity": "sha1-HZiYEJVjB4dQ9JlKlZ5lTYdqy/U="
     },
     "supercluster": {
       "version": "7.1.0",
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "@types/node": "^14.14.7",
     "@types/react": "^16.9.56",
     "@types/react-dom": "^16.9.9",
+    "@types/suncalc": "^1.8.0",
     "bootstrap": "^4.5.3",
     "d3": "^6.2.0",
     "ecef-projector": "^1.0.1",
@@ -25,6 +26,7 @@
     "react-leaflet": "^2.6.3",
     "react-plotly.js": "^2.5.0",
     "react-scripts": "^4.0.0",
+    "suncalc": "^1.8.0",
     "typescript": "^4.0.5"
   },
   "scripts": {
--- a/src/App.js
+++ b/src/App.js
@@ -37,6 +37,7 @@ import {
   CalculatorPanel,
   //solar tab
   Boron8SpectraPlot,
+  AnalemmaPlot,
 
   // Custom Core UI
   AddCustomCoreModal,
@@ -250,6 +251,7 @@ function App(props) {
               </Tab>
               <Tab eventKey="solarnu" title="SolarNu">
                 <Visible>
+                  <AnalemmaPlot detector={detector} cores={cores}/>
                   <Boron8SpectraPlot boron8={boron8} />
                 </Visible>
               </Tab>
--- a/src/App.js
+++ b/src/App.js
@@ -251,7 +251,7 @@ function App(props) {
               </Tab>
               <Tab eventKey="solarnu" title="SolarNu">
                 <Visible>
-                  <AnalemmaPlot detector={detector} cores={cores}/>
+                  <AnalemmaPlot detector={detector} cores={cores} reactorLF={reactorLF}/>
                   <Boron8SpectraPlot boron8={boron8} />
                 </Visible>
               </Tab>
--- a/src/detectors/index.ts
+++ b/src/detectors/index.ts
@@ -1,4 +1,5 @@
 import {zip} from 'lodash';
+import SunCalc from 'suncalc';
 
 type DetecorPreset = {
     name: string,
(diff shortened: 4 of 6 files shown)


stdye:
It is great to see progress on this issue!

stdye:
Looks better. It is not clear what quantities are plotted on the axes. Check out the plot for SUPL in the southern hemisphere

stdye:
Nice. 

Questions: Do you plan to

1) Force azimuth axis tick values to be multiples of 15 deg (i.e. 45, 90, 135, 180, etc.)?

2) Color-code and add hover labels to the reactor core dots?

stdye:
Oh, very nice.

The analemma curves should tie together with the selected reactor LF date range (e.g. Only half of each hourly analemma displays if the date range if for just 6 months).

Have you looked at the plots for detector locations: INO (near equator) and SUPL, Andes (southern hemisphere)?

Do the reactor positions need adjusting for southern hemisphere detector locations?

Have you conquered the problem of the output file containing the directional data?

DocOtak:
Responses to questions:

* It shouldn't' be too hard to tie this to the LF range.
* I have looked at them, things appear correct. I found myself looking at a bunch of [antipode maps](https://en.wikipedia.org/wiki/Antipodes) when trying to "wrap" my head around if the core locations seem to be right.
* I don't think so, almost all the cores when viewed from the southern hemisphere will be really "down".
* Directional data for cores? Because solar seems "hard", either one I haven't given much thought to yet.

stdye:
Very good. 

It makes sense to me with realization that the sun travels CW in azimuth in the northern hemisphere and CCW in azimuth in the southern hemisphere.

For analyzing directional signals from reactors (IBD and ES) one can always download the individual core rate spectra and note the theta and phi values as given by the big core list. 

Note: we changed azimuth values in the core direction plot but maybe not the azimuth (phi) values in the core list?

For analyzing directional signals from the sun (ES only) one needs the rate (spectrum should always be the same) for each altitude-azimuth (theta-phi) direction. One task is figuring out how to store all the data in the smallest possible file size.

stdye:
I think this can be merged once the analemma curves are tied to the reactor LF date range. What do you think?

stdye:
Super!

stdye:
Why are the bottoms of the analemma loops missing?

stdye:
Bottoms appear when starting LF range at 2018-12 and end LF at 2019-12. This is 13 months though. Maybe currently the curves start at the end of 2019-01 rather than the start of 2019-01?

stdye:
Line 32 in solar-plot.jsx

let startMonth = (reactorLF.start.getUTCMonth() + 1).toString().padStart(2, "0")

Why + 1? Doesn't this put the start at the end of the month?

stdye:
One task and one question?

Task: Make an appropriate portion of the analemma loop appear on the plot when the start and end month are 01 (January).

Question: Are the analemma loops always for the year 2021?

stdye:
This fixed 2019-01 to 2019-01 but now 2019-02 to 2019-02 shows the almost full loops.

DocOtak:
Yay browser specific implementation bugs.
The +1 (and now +2) for the month number is not intuitive at all because months are [zero indexed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCMonth) in JavaScript. This means that January is 0 and December is 11.

stdye:
This looks good.

Because Earth's orbit is not exactly circular, the solar neutrino flux at Earth varies over the year with an amplitude of about 3.5%. The flux is highest at perihelion (~Jan 4) and lowest at aphelion (~Jul 4).

We want to include this flux variation, which is simply a 1/r^2 factor, on the existing analemma plot and in the solar data output file.

Do you want the inclusion of flux variation to be a separate issue on GitHub or can you tackle it in this issue?

DocOtak:
What would this distance parameter look like on the analemma plot? The only thing I can think of is some sort of coloring of the analemma itself.

stdye:
Yes. Color varying smoothly within a loop would be perfect. I tried to do this on the plots in the paper but my knowledge of Root did not let me do this easily. So I just used different colors for the different seasons but this is pretty lame.

stdye:
The flux variation factor should be normalized to 1, so values would be between ~0.965 and ~1.035. The value 1 would ideally correspond to the time-average Earth-Sun distance. In my first attempt to get this right, I used the semi-major axis a as the normalizing distance with the flux variation factor given by (a/r)^2.

stdye:
It appears that SunCalc- https://www.npmjs.com/package/suncalc -does not return Earth-Sun distance for each calculated position (it does return the Earth-Moon distance). Is this the package we are using? If so, what is the plan for getting the Earth-Sun distance as a function of time (orbit)?

DocOtak:
Suncalc doesn't provide distance. But the Kepler first law equation seems pretty simple to implement for a good approximation using Jan 3 or 4 for perihelion.

stdye:
It looks pretty simple to implement Kepler's first law by finding position as a function of angle r(psi) as in Eq 50. However, we want the position as a function of time r(t) rather than r(psi). Converting r(psi) to r(t) I think is not so straightforward; it requires the function psi(t), no? What is your plan?

DocOtak:
Hm... Looks like we have options we can calculate the [Equation of center](https://www.aa.quae.nl/en/reken/zonpositie.html#3) to put into the Kepler's first law equation. Though I'm tempted to try [this from wikipedia](https://en.wikipedia.org/wiki/Position_of_the_Sun#Ecliptic_coordinates).

stdye:
My code uses the equation of center-

c Julian century
         cjul=(djul-djul0)/36525.d0
c geometric mean longitude sun (deg)
         gmls=mod(gmls0+cjul*(36000.76983d0+cjul*0.0003032d0),360.d0)
c geometric mean anomaly sun (deg)
         gmas=gmas0+cjul*(35999.05029d0-cjul*0.0001537d0)
c eccentricity earth orbit
         eccn=eccn0-cjul*(0.000042307d0+cjul*0.0000001267d0)
c sun equation of center
         eqoc=sin(gmas*rddg)*(1.914602d0-cjul*(0.004817d0+
     +cjul*0.0000014d0))+sin(2.d0*gmas*rddg)*(0.019993d0-
     +cjul*0.000101d0)+sin(3.d0*gmas*rddg)*0.000289d0
c sun true anomaly
         stra=gmas+eqoc
c sun radial vector (a.u.)
         srve=1.000001018d0*(1.d0-eccn*eccn)
         srve=srve/(1.d0+eccn*cos(stra*rddg))
c solar nu flux variation
         flxv=1.d0/srve/srve

stdye:
Hovering on the analemma curves shows the altitude-azimuth coordinates.

Additional data- date and time, flux variation factor- would be helpful. For example, one could cross check the alt-azi values against other calculations.

stdye:
Very nice!

stdye:
Solar neutrino flux and interaction rate at Earth scale with Earth-Sun distance^{-2}. Assuming the Super-K time-averaged flux  applies to 1 a.u., then the 1/d^2 factor would be good on the z-axis. This would make the closer distances plot in red (higher flux- hotter?) and the farther distances plot in blue (lower flux- colder?).

stdye:
Trial and error. Want to cross check azimuth, altitude, and distance values for a given date and detector location. Couldn't get them to display on the plot.

https://plotly.com/javascript/colorscales/

The examples at this link make it look like x, y, z values are shown by default. 

Maybe just removed line 113 in solar-plots.jsx?

stdye:
This version of the routine for the analemma plot produces hourly curves, which are color-coded for Earth-Sun distance, for a given detector location for a selected reactor year/month range. If you are confident that the curves are correct, you may merge this pull request.

Estimating the solar neutrino rate (ES NIUs) versus (cosine of the) angle from the vector between a detector and a given reactor core can be covered by a new issue.

DocOtak:
I'll get this merged in tomorrow

PR #180, opened 2021-03-08
Update App.js

stdye:


stdye:
minor reformatting of Output page and improved link to reference paper

PR #183, opened 2021-03-09
Update plot.jsx

stdye:
minor changes to text and title


stdye:
The text at the bottom right of the plot (geoneutrinos.org label and definitions of the plot axes) gets cut off on the right in the downloaded plot (Safari browser).

PR #184, opened 2021-03-09
Update plot.jsx

stdye:


stdye:
Adding ' &nbsp; &nbsp; ' before each '<br />' pushed the text over to the left enough to just barely see the full string on the downloaded plot using Preview with Safari browser.

PR #185, opened 2021-03-10
Update solar-plots.jsx

stdye:
try adding geoneutrinos.org label to solar analemma plot


stdye:
Placing the geo neutrinos.org label outside the plotted data area was needed for southern hemisphere locations (e.g. Andes)

PR #193, opened 2021-03-10
Update plot.jsx

stdye:


stdye:
minor changes to legend titles on main plot

PR #200, opened 2021-03-10
Update reactors-corelist.jsx

stdye:


stdye:
Changed Azimuth and Elevation labels, added degree symbols

PR #203, opened 2021-03-14
Add core sort by intensity to reactors-corelist.jsx

stdye:
Start to add Core list sort by intensity (Thermal Capacity / Distance^2). See if change adds "Sort By: Intensity" to the pull down list. If it works, can go from there to add code to do the sort. It might fail b/c "intensity" may need to be defined. Not sure. One way to find out is to try.

### DIFF
--- a/src/ui/reactors-corelist.jsx
+++ b/src/ui/reactors-corelist.jsx
@@ -110,7 +110,8 @@ const CoreListItem = ({
           </span>{" "}
           {core.power} MW<sub>Th</sub>
           <br />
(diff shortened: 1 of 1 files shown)


stdye:
Haha! "Intensity" appears on the pull down list. Moreover, the code is pretty robust because selecting "Sort By: Intensity" does not crash the site, the sort defaults to "Sort By: Name"

stdye:
This can be achieved by setting all cores to maximum power and then sorting by signal. Like sort by angle this is a diversion. We have important issues associated with the elastic scattering signal to work on.

Issue #208, opened 2021-07-02
Rate Significance Calculator on Output Tab- upgrade 


stdye:
Two restrictive simplifications make the "calculator" less useful than it could be.

### 1) The significance calculation considers only **statistical** uncertainty; It ignores **systematic** uncertainty on the background.

Because the antineutrino backgrounds are estimated, the significance of detecting the signal is degraded by the systematic uncertainty of these estimates. A better expression for the significance is 

(Written originally in this comment as

N_sigma = Signal x sqrt(Time) / sqrt(Signal + Background + deltaBackground^2),

but see comments below for corrected equation)

where deltaBackground^2 is the sum of the squares of the systematic uncertainties of the rates of the various backgrounds.

Default values of the fractional systematic uncertainties of the rates would be 25% for geo-nus and 6% for reactor antinus. Users should have the ability to set the quantities to different values. 

### 2) Results apply only to a detector with 10^32 targets.

Users should enter the number of targets as a factor of 10^32. The significance (N_sigma) is then 

N_sigma = Signal * sqrt(Time x **Factor**) / sqrt(Signal + Background + deltaBackground^2). 

In the meanwhile, a note that tells users to account for this factor needs to be added to the Rate Significance Calculator Pane. 

**Note:** Coding up a new variable for the exposure factor can be avoided if the "Solve For" option is simply "Exposure" rather than "Exposure Time". In this case the value-setting box now called "Exposure Time for 10^32 Targets" with unit "years" becomes simply "Exposure" with unit "10^32 target-years". Users can easily solve for the time using the number of targets for their detector.

stdye:
CAUTION:

N_sigma = Signal x sqrt(Time) / sqrt(Signal + Background + detlaBackground^2) 

is NOT correct. The quantities in the radical in the denominator have different dimensions. Mea culpa!

The following equation is correct

N_sigma = S / sqrt(Signal + B + detlaB^2),

where S is the NUMBER, not the rate, of signal events and B is the NUMBER, not the rate, of background events,.

S = Signal (NIU) x Factor x Time (yr) 

B = Background (NIU) x Factor x Time (yr)

deltaB = B x deltaB/B

DocOtak:
Currently trying to grok this one for implementation...

I did a little test just solving for significance:
Nσ = (Signal * Time) / sqrt(Signal + Background * Time)

This results in the significance rising faster than the current equation (as a function of exposure time)

For the uncertainties:
[...]

Issue #199, opened 2021-03-10
Add Geo 235U signal


stdye:
Implementation would be very similar to Geo 40K; It only shows up for ES cross section options. Color for filled area on main plot should be the same color (currently purple) as the 235U line on the Geoneutrino Spectra plot.

235U gets scaled by the 238U value, getting knocked down by the low natural abundance but pushed up by the higher isotopic luminosity due to the relatively short half-life.

I expect that the 235U signal is quite small but we have all the required data already in the web app. Why not include it?

stdye:
The 235U flux relates to the 238U flux according to Eq. 41 in our paper on arXiv. Solving for 235U flux gives

phi_235U = phi_238U * [ l(235U) * NA(235U)] / [ l(238U) * NA(238U)]

The isotopic luminosities of 238U and 235U as listed in Table XI of our paper on arXiv are 74.601 /kg/us and 319.841 /kg/us, respectively.

The natural abundances (%) of 238U and 235U as listed in Table XIII of our paper on arXiv are 99.2745 and 0.7200, respectively.

Adding the Geo 235U signal is very straightforward.

PR #210, opened 2021-07-16
Use background uncertainties for output calculator

DocOtak:
Also includes mathjax for nice equations, the UI for changing the
uncertainty and target scale isn't done yet.

Tracking Issue #208


DocOtak:
This obviously isn't ready to go in yet, but I'd like some sanity checks before I put the rest of the UI elements in. 3 new numeric input boxes: reactor core uncertainty (%), geonu uncertainty, and exposure scale factor.

stdye:
Coding up a new variable for the  exposure factor can be avoided if the "Solve For" option is simply "Exposure" rather than "Exposure Time". In this case the value-setting box now called "Exposure Time for 10^32 Targets" with unit "years" becomes simply "Exposure" with unit "10^32 target-years". Users can easily solve for the time using the number of targets for their detector. 

BTW- I think the label for the N_sigma value-setting box should not have "for 10^32 Targets".

stdye:
The Calculator doesn't work right when the signal is from the Custom Core

DocOtak:
Oh... you got to it before I did

stdye:
It was an easy fix. I was using a Custom core at the Hartlepool site to make the signal from the Hartlepool complex rather than the using closest core. Entering the lat/lon/elev values is a bit tedious but it works. It would be nice if the suggested custom cores menu allowed users to select sites from the list of existing IAEA cores. Of course, a more direct solution lets users select the signal from a complex of cores.  

BTW- Marc and Steven at Livermore are reviewing the upgraded calculator on the development page.

stdye:
"Infinite" is much better than 999999999.999 in the exposure box when N_sigma * background_uncertainty > signal. However, it triggers the isInvalid logical N_events < 2 when it should not.

DocOtak:
Mathematically, "infinity" in that function is actually negative times, so I put in a check for that.

stdye:
The implementation of the Nuisance Background is blind to the IBD vs ES cross section; It adds background for E_nuebar < 1.806 MeV (the IBD threshold energy). This is fine for ES but not right for IBD.

It looks like output-calculator.jsx doesn't know which cross section produces the signal. If this information gets imported, it can be used to apply the nuisance background more appropriately.

stdye:
import {PhysicsContext} from '../state' ???

DocOtak:
Importing and using the physics context is how the calculator could get access to the current XS function, that context is a little complex, do we need the current function itself (will it be called?) or just the name of the function (which we would use in an if/else statemnet)?

stdye:
just the name of the function, which we would use in an if/else statement, would be just fine

stdye:
If you are changing the code to pass the cross section name, could figure out how to add a space between "10^32" and "target-years" in the unit displayed on the Exposure value-setting box? I tried a few things but none that worked.

DocOtak:
Would the... logical place to change this to be setting "Antineutrino Emin" min to the IBD threshold if one of the IBD functions is currently in use?

stdye:
That sounds right. The code now has a "magic" number (0.1 = 1/10) that it uses to scale the background. This number should be calculated from eMax and eMin

DocOtak:
This line?
```javascript
const bkgNuisanceNIU = bkgnuisance * (eMax - eMin) * 0.1
```

I think it would stay the same, we would just ensure that eMin is >= IBD threshold energy, unless this background needs to be calculated some other way.

stdye:
That's the line. The idea is to scale the nuisance background, which is assumed to have a flat energy spectrum.

For ES-
const bkgNuisanceNIU = bkgnuisance * (eMax - eMin) / (10 - 0)

For IBD-
const bkgNuisanceNIU = bkgnuisance * (eMax - eMin) / (10 - 1.8)

stdye:
The lower limit of the Antineutrino Energy in the settable-value box depends on the cross section. For IBD it is E_thresh for ES it is 0. 

const UIsetEMin = (event) => {
    const value = event.target.value;
    let e_min = parseFloat(value);
    if (isNaN(e_min)) {
      setEMin(value);
    } else {
      if (e_min < 0) {
        e_min = 0;
      }
      if (e_min > 10) {
        e_min = 10;
      }
      if (eMax < e_min) {
        setEMax(e_min);
      }
      setEMin(e_min);
    }
  };

DocOtak:
You also want to set the default value (the argument to `useState`) to be what we expect for the initial state of the app, which starts in IBD mode.

DocOtak:
Do we want a UI for the the user to set uncertainties (the delta* variables)?

Some reasonable min/max/step values for each would be useful for this.

stdye:
Ah, yes, useState. I will change that next. The constant for the threshold energy was not acceptable; It's value got stuck in the E_min box.

Eventually we may want UIs for setting the deltas. The note at the bottom tells the user the delta values, which are reasonable.

DocOtak:
Let me know when you think we are close to merging, I'll need to fix the thing causing the netlify build to fail.

stdye:
Not sure what to try next to keep E_thresh value from getting stuck in E_min box

stdye:
We are getting close but there are still several things to do before merging-

1) The web app now allows consideration of the signal from the closest core but not the signal from all the cores within the closest complex. For nuclear reactor monitoring demonstrations, there is currently much interest in estimating the exposure required to detect the signal from all the cores within a given reactor complex (i.e. Both Hartlepool cores as the signal with geo-nu + other cores as the background;  And/Or with both Hartlepool cores off, all 4 cores in the Heysham complex as the signal with geo-nu + other cores as the background). One can effect this by adding a custom core that burns the fuel mix with the summed power of all the cores within a complex and putting the custom core at the location of the complex and then turning off the cores in the complex. This is a pretty tedious process. There must be an easier way to do this.

2) Although the code executes and things "look" reasonable, we need to verify the results quantitatively. In fact, users would benefit, and have more confidence in the calculator, from being able to see the numbers (rate * exposure) of events from the various signal and background sources. These numbers are now hidden from the user. A simple dashboard with displayed number tallies would be great and help us to verify the results.

3) Add a note somewhere on the Calculator that indicates which cross section is being used: IBD or ES.

The following is an important upgrade but it could be implemented later. 

4) Energy-dependent detection efficiency functions.

stdye:
output-calculator.jsx almost fully adapted for ES signals from geo-40K and geo-235U.

Making the Signal (background) options conditional on the cross section (IBD vs ES) was not attempted. Ideas, sensei?

Would be very interesting to have geo-235U implemented. See Issues #199 and PR #211

stdye:
Also not yet implemented is the assignment of the uncertainties to the ES rates at E_nuebar < 1.8 MeV. These uncertainties are definitely larger than those at E_nuebar < 1.8 MeV but how much larger? Maybe the code needs to separately account for the ES rates above and below E_nuebar = 1.8 MeV.

DocOtak:
Thoughts:
1. I think for this we need to change the reactor core data model a bit, while not explicitly stated, I think all the core name follow a name-# where name is the "complex" and -# is some individual core. I think this grouping should be done at the database ingest stage (the python scripts which turn the excel into json). And while in there, add a slot of permanent shutdown date information for that _other_ problem we have.
2. Dashboards are easy, we should use that "clickable" to swap to full precision element so we can verify the "exact" calculations.
3. Notes are easy too.

----

If I recall correctly, <sup>40</sup>K and <sup>235</sup>U are only seen in ES interactions so we would not want them as options if the interaction is IBD.

stdye:
started coding for reactor signal and uncertainty at energy less than E_thresh, comments added where code needs development

please activate geo U235

stdye:
Adding detection efficiency (as requested by Marc B) as a function of antineutrino energy E (only for IBD, for ES the function depends on the kinetic energy of the scattered electron T and can be added later).

Three input parameters (default values): E_start (E_thresh), epsilon_saturation (0.8), ramp_up (1)

epsilon(E) = epsilon_saturation * (1 - exp(-ramp * (E - E_start))) E>=E_start

Try it on a web-based graphing calculator- https://www.desmos.com/calculator

This function gets multiplied by the rate spectrum before summing for the total. Otherwise everything stays the same.

Easy, eh? Let's do it!

Questions?

stdye:
Detection efficiency function gets applied to the antineutrino signal and antineutrino background rate spectrums but NOT the nuisance background

DocOtak:
Hm...
https://www.desmos.com/calculator/sihazxmzrt

stdye:
Yes! Only you missed the minus sign before the ramp_up variable

DocOtak:
https://www.desmos.com/calculator/a0tsxacry3

Is the desire to have this as a plot on the calculator page?

stdye:
That's a good idea. The detection efficiency equation is sufficient, though. Physicists will recognize the equation; It has the same form as the charge on a capacitor as a function of time as the capacitor charges through a resistor.

q(t) = Q * (1 - exp(-t/RC))

PR #211, opened 2021-07-16
Update detector-stats.jsx

stdye:
Starting to add 235U geo-neutrino rates to Spectrum Stats pane

### DIFF
--- a/src/ui/detector-stats.jsx
+++ b/src/ui/detector-stats.jsx
@@ -7,6 +7,7 @@ import { Elements } from "./elements";
 import {
   rateToFlux232Th,
   rateToFlux238U,
+  rateToFlux235U,
   rateToFlux40K,
 } from "../antineutrino-spectrum";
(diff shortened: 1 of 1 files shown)


stdye:
This obviously is not ready to deploy but it does successfully add a variable quantity for the 235U signal with a dummy value of 0.0 NIU that shows on the Spectrum Stats pane when ES cross sections are selected. Now to figure out how to make the dummy into a smarty.

stdye:
Unfortunately, the existing code is largely devoid of comments. One needs to already know JavaScript or try to figure it out on the fly.

stdye:
Barna- Put in some effort to help this along. All the parts are already in the code, they just need to be gathered and assembled. See Issue #199

DocOtak:
This should be done before the calculator improvements (so the calculator can use the data). I'm going to make the <sup>235</sup>U use the same code paths as the rest of the geonu so there are a bunch of places that need changing.

PR #212, opened 2021-07-21
Add 235U geo signal

DocOtak:
Work in progress PR for adding the 235U signal.


DocOtak:
Right now this adds a <sup>235</sup>U geonu calculation, there is a temporary line on the stats pane to see the NIU signal, and it has been added to the main plot. There is only a signal when an ES interaction is selected and you'll need to zoom way into the plot to see the spectrum (but it's there).

This currently assumes the crust lookup is entirely <sup>238</sup>U. (The mantle is explicitly <sup>238</sup>U on the page)

stdye:
Looks good! According to Huang and McDonough, crust lookup U flux is entirely U238

DocOtak:
Before making a lot of cosmetic changes, does the <sup>235</sup>U calculation appear to be correct?

DocOtak:
Next steps will be to include this <sup>235</sup>U signal in all the places that will want to use it (geo totals, etc..).

DocOtak:
Please advise on if the (Th/U) and (K/U) should be for 238 + 235 as the "U", right now it is only using 238, I've updated the label to be specific about this.

stdye:
Geo 235U rate spectrum looks to be ~x25 less than 238U as expected. The new code to scale the 235U spectrum look correct to me.

stdye:
Th/U and K/U mass abundance ratios of the elements (not the isotopes) are calculated with Eq 46 and 47, respectively.

stdye:
I looked through the code at all the calculations for luminosities and element mass abundance ratios, etc. and they correspond to the equations in the paper. I did not check the actual values of the constants used in the equations.

The labels for the mass abundance ratios posted on the stats panel should **_not_** specify isotopes; They are for the elements.

Of course, because the NA of 238U is almost = 1 (0.993), there is little difference in the values.

DocOtak:
The code path that generates the Th/U and K/U ratios has some <sup>238</sup>U specific stuff going on, options:

1. leave existing code path in place, accept the small "error" from having a very low <sup>235</sup>U signal, remove <sup>238</sup>U label from both Th/U and K/U.
2.  Account for the <sup>235</sup>U signal using the appropriate rate to flux for each isotope then summing the U fluxes before taking the Th/U and K/U ratios, remove <sup>238</sup>U label from both Th/U and K/U.
3. something else?

stdye:
Leave the code for calculating Th/U and K/U as it is. It uses Eq. 46 and 47 from the arXiv paper. 

Remove the 238U label from both Th/U and K/U.

The ratios as calculated are for the mass abundances of the elements.

DocOtak:
I think I've found all the places where <sup>235</sup>U should be used. Once we merge this, we/U can update the new calculator to use the "real" signal.

stdye:
a_238U = a_U * NA_238U; a_235U = a_U * NA_235U; a_U = a_238U + a_235U = a_U * (NA_238U + NA_235U)
[the last equation above ignores many uranium isotopes but they all have extremely small NA]
 
a_K = a_40K / NA_40K

stdye:
Since the web app calculations currently consider only 40K_beta and ignore 40K_EC, why not add the beta subscript? This might avoid confusion in the future.

Although the electron capture (EC) decay of 40K has a branching ratio of almost 11%, they make a neutrino (not an antineutrino) that rarely has more energy than 43 keV.

stdye:
Also on the Spectrum Stats pane for R_geo

stdye:
This looks very good! Merge at will.

DocOtak:
Ok merged, the calculator work in progress now "conflicts" will need to be reworked a little before it again.

PR #219, opened 2021-07-27
Update output-calculator.jsx

stdye:
starting upgrade- change rate and exposure symbols; italicize antineutrino energy symbol E; try to put E_max and E_min controls in two columns on same row


stdye:
This feels like a waste of time...

PR #220, opened 2021-07-27
Update output-calculator.jsx

stdye:
Try Cloudflare build on unedited output-calculator.jsx. Only added a new line with an empty comment

### DIFF
--- a/src/physics/neutrino-oscillation.ts
+++ b/src/physics/neutrino-oscillation.ts
@@ -75,7 +75,7 @@ export let oscillation: Oscillation = {
   massOrdering: MassOrdering.Normal
 }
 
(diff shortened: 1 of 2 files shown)


stdye:
So it can work... painfully slowly, however

stdye:
so placeholder "useState" till there but options removed

DocOtak:
"Prettier" is [something specific](https://prettier.io/)

DocOtak:
Re set* methods, made by `useState`. The useState function returns an array that we are [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). that second element needs to go somewhere even if it isn't used (really need to just add the UI to actually allow editing).

PR #223, opened 2021-07-29
Disable detectorEfficiency if interaction type is ES

DocOtak:


stdye:
I am checking things out. So far one small problem-

When selecting an ES cross section, Antineutrino E_min comes up as E_thr rather than 0

stdye:
The previous comment identifies a legacy problem. It does not appear to have been introduced by this PR

stdye:
The signal and background events check out when solving for "significance", setting exposure to 1x10^32 target-years, and selecting Signal (background) to Geoneutrinos (reactors). For IBD, the maximum detection efficiency needs to be set to 1.0 and the ramp_up parameter to about 1000. With these settings the numbers of events agree with the rates in NIU given on the Detector tab.

stdye:
The Antineutrino E_min control does not change the numbers of events as expected.

Raising E_min above E_thr doesn't change the Reactor events. It does change the Geonu events as expected.

Lowering E_min below E_thr does change both the Reactor and the Geonu events as expected.

This aberrant behavior also looks like a legacy issue; It is present in the current live version of the web app. I do not recall this problem in the version prior to launching the detection efficiency but maybe I missed it.

stdye:
Looks very good! Merge at will

Issue #242, opened 2021-08-04
Corelist- Add "Signal" Button


stdye:
Give users more ease and flexibility in selecting which cores contribute to the Signal used by the output calculator-

Idea: Add a "Signal" checkbox on each core in the core list. The checkbox controls a boolean variable for "Signal" that is initially false but is set to true by checking the box. The output calculator treats the aggregate of all checked cores as the signal and all unchecked cores as background. It seems simple to me but I do not fully appreciate the architecture of the code.

There should probably be an option on the "Control all cores" menu that resets all cores to background.

**What do you think?**

This would simplify the process of selecting the signal from a reactor complex and add the possibility of selecting any combination of cores (i.e. two cores from Heysham plus two cores from Torness; all the cores from multiple complexes like Bruce, Pickering, Darlington in Canada, which would take care of #225 ). Continuing to use the workaround by adding new reactor complexes to the Custom Core Location Suggestions menu is not very practical. It is a kludge in the code. No?

stdye:
What are your thoughts on this issue?

Your response is requested when you get a free minute.

DocOtak:
Closed with #243

PR #243, opened 2021-08-06
Add "use as signal" checkbox for use in output calculator

DocOtak:
This is an early implementation/test of a way to handle #242

It probably needs a little more polish/features before merge. Please play with it and let me know if this looks like something that will address #242.

### DIFF
--- a/src/reactor-cores/index.ts
+++ b/src/reactor-cores/index.ts
@@ -227,6 +227,7 @@ interface ReactorCore {
   detectorNIU: number;
   direction: Direction;
   powerFractions: PowerFractions;
+  outputSignal: boolean;
 
   setSignal: (
     dist: number,
@@ -380,6 +381,7 @@ export function ReactorCore({
     setSignal: setSignal,
     direction: { phi: 0, elev: 0 },
     cos: cos,
+    outputSignal: false,
   };
 }
 
--- a/src/ui/output-calculator.jsx
+++ b/src/ui/output-calculator.jsx
@@ -235,7 +235,7 @@ export const CalculatorPanel = ({ cores, spectrum }) => {
     getCoreSums(coreList, min_i, max_i, low_i);
 
   const selectedCores = coreList.filter(core => core.outputSignal)
-  const [selctedCoreSignal, selectedCoreSignalLow, selectedCoreSignalHigh] =
+  const [selectedCoreSignal, selectedCoreSignalLow, selectedCoreSignalHigh] =
     getCoreSums(selectedCores, min_i, max_i, low_i);
 
   // custom cores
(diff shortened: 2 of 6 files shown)


stdye:
This will address #242 .

1) An option on the "Control all Cores" to clear all the checked Signal boxes would be a convenience to users

2) It would be super nice to see the selected Signal spectrum on the main plot, or maybe a new plot on the Output tab only

stdye:
To facilitate analyses of the Signal (background) spectra, users will be asking for a download option

DocOtak:
In playing around, it seems that a "Select all Filtered" cores might be very desirable. e.g. if I filter the core list for "BRUCE", I'll probably want to select all 8 cores at the complex as my signal....

stdye:
Indeed, "Select all Filtered" would be desirable

stdye:
Feel free to remove the "complex" selections I recently added to the Custom Core- Core Locations Suggestions and push it to this signal_box branch. They are a kludge but I was under a deadline to get something working.

stdye:
Okay, Houston, we have a problem here. 

1) There is a 235U column on the GeoNu output file. Probably a legacy problem from adding 235U signal.

2) Output file title shows as "...Tminx.xMeV.csv" but T_min not active for IBD. Probably also a legacy problem from adding T_min to ES.

Otherwise, this looks good to me.

stdye:
I suggest we remove "Tminx.xMeV" from IBD filenames until T_min is activated for IBD

DocOtak:
For the core types change, if you type one of the core types "exactly" (case insensitive) into the reactor list filter box, it will switch to filtering on that rather than the core names.

This is a "quick" solution that will allow setting the "signal" to be all cores of a particular reactor type (e.g. PHWR -> all the CANDU cores)

stdye:
Very nice... I only had trouble selecting MOX

stdye:
Legacy "MeV" at end of IBD filename

DocOtak:
The magic term "LEU_MOX" should work now (once the preview build finishes).

stdye:
Very nice!

Small legacy issue- Reactor download filename starts with "AntiNu". It should probably be changed to start with "Reactor" or something similar

Somewhat stale (you were at sea at the time) user request that came to me by email from Slava at Livermore- He liked both reactor + geonu in one file. The filename for this appropriately starts with "AntiNu". I guess the button label would be "Reactor + GeoNu". He complained that he needed to change his code for reading the files. I am sure he has done this by now so there is really no need to (re)make another download. Just an FYI b/c it relates to the "AntiNu" legacy above

DocOtak:
Merge time? Probably also send some updated instructions to the collaboration?

stdye:
I am not sure that all the signal events (S) are showing in the output-calculator.

I set Detector at SNOLab and filter-selected PWR (330 cores) and set efficiency params to 100%. The signal events (S) came to less than 12 but the rate curve on the plot looked to have an average of about 10 /MeV over a 5 MeV range. I was guessing that there would be maybe 40 or 50 signal events

stdye:
My stupid. I forgot to change the Signal (background) option from Closest (geoNu + other reactors)

stdye:
Merge at will... I will send email to known users

PR #244, opened 2021-08-07
rename some backgrounds, set default to "selected"

stdye:


DocOtak:
Be aware that this will start off with no signal by default.

stdye:
Yep, starting off at 0 signal leads users to the cool new core selection features

PR #247, opened 2021-08-09
trial and error- change core circle size

stdye:
radius from 1000 to 2000


stdye:
Only changed core circle size when zoomed all the way in so this is not what Adam wanted

some discussion of this topic on the web-
https://stackoverflow.com/questions/22467177/draw-a-circle-of-constant-size-for-all-zoom-levels-leaflet-js

stdye:
again this changes the size of the detector circle only when zoomed all the way in...

there must be a minimum circle radius rendered on the map when zoomed out... now look for this

stdye:
going to close this trial and error PR and start an issue

Issue #248, opened 2021-08-09
Circle marker size on Leaflet map


stdye:
We have a request from user @Adam to make circle markers for cores and detectors larger when zoomed in but not all the way in (e.g. set zoom so that only England, Scotland, and Ireland show on the map- the circle markers are very hard to see because they are very small). User wants to take map screenshots to drop into presentations.

The two circle radius = value instances in ...src/ui/map.jsx seem to only control the size of the detector and core circles when zoomed all the way in. Maybe there is a minimum circle size set in .../src 

Upgrade ideas to consider...

1) Make area of circle proportional to core thermal capacity
2) Overlay a marker to designate that the core has been permanently shutdown

DocOtak:
There are two types of circle markers:
CircleMarker has a size specified in pixels: https://leafletjs.com/reference-1.7.1.html#circlemarker
Circle (what we are using currently) has size specified in meters: https://leafletjs.com/reference-1.7.1.html#circle

What we might end up needing to do is listen for the map "zoom" event and scale the markers accordingly.

stdye:
Leaflet documentation is not great for the novice. Circle rather than CircleMarker looks fine. I was just reading this...

https://gis.stackexchange.com/questions/203574/lock-circlemarker-size-leaflet-r

PR #249, opened 2021-08-09
change map core circle size based on zoom

DocOtak:
This is a start at a possible solution for #248 

Right now there are two break points for zoom levels and it only operates on the core objects.

I'm guessing this will need some tuning.

### DIFF
--- a/src/ui/map.jsx
+++ b/src/ui/map.jsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import {
   Map,
   Popup,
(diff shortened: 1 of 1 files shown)


stdye:
It looks like this will work but can be improved with tuning, like you say.

DocOtak:
Any feel for what the breakpoints/marker sizes should be?

stdye:
It looks good zooming in. Only the final jump looks too big to me.

stdye:
Added more steps and adjusted the values... looks smoother... could add more steps or maybe make a continuous function... what is your plan for the detector circle?

Issue #201, opened 2021-03-11
Make Azimuth values in CoreList agree with Azimuth values on Analemma plot


stdye:
The azimuth values in reactors-corelist.jsx are on line 
124  "Azim: {core.direction.phi.toFixed(1)}&deg;"

The azimuth values in solar-plots.jsx are modified on lines 17-19 (we want the modified values on the core list)
17     .map((v) => -(v.direction.phi * (Math.PI / 180)) + Math.PI / 2)
18     .map((v) => (v < 0 ? v + 2 * Math.PI : v))
19     .map((v) => (v * 180) / Math.PI),

This makes core.direction.phi have a value different from core azimuth angle

E.g. for Hartlepool cores as seen from Boulby azimuth is given as -157.8 deg on the core list but azimuth is given as ~290 deg (value not given by hover pop-up but should be 292.2 deg) on the analemma plot

Core-list azimuth value should be the same as the azimuth value on the analemma plot for consistency and to facilitate direction studies by users.

stdye:
**Problem:** The azimuth values shown on the Core List use the ENU convention, which is different from the NED convention that we use in the Core Directions at Detector bullseye plot on the Detectors tab and in the Solar Analemma plot on the SolarNu tab. Moreover, our arXiv-posted paper uses azimuth as defined in the NED convention not the ENU convention. 

**The Fix?:** Reverse the arguments of the atan2 function on line 74 in .../src/detectors/index.ts 

74    phi:atan2(enu.y, enu.x) * (180/Math.PI),

to become

74    phi:atan2(enu.x, enu.y) * (180/Math.PI),

Wouldn't this convert azimuth from counterclockwise from East (ENU- reactor core list values) to clockwise from North (NED- Solar Analemma plot values and Core Directions at Detector bullseye plot)?

This change should allow lines 16-19 in .../src/ui/solar-plots.jsx 

16    x: cores
17    .map((v) => -(v.direction.phi * (Math.PI / 180)) + Math.PI / 2)
18    .map((v) => (v < 0 ? v + 2 * Math.PI : v))
19    .map((v) => (v * 180) / Math.PI),

to become simply 

16   x: cores.map((v) => v.direction.phi),

and lines 201-202 in .../src/ui/plot.jsx

201      tickvals: [0, 45, 90, 135, 180, 225, 270, 315],
[...]

PR #253, opened 2021-08-13
Include shutdown date in core database

DocOtak:
This is another work in progress PR to fix up the core database including shutdown dates. Currently it just displays if a core has been shutdown prior to "today".


stdye:
Is this ready to merge? Please see last comment on Issue #201

stdye:
How about remove the "Caution: ..." note and change the drop down option to "Use 100% LF for all active cores" ?

It is nice that the individual core override option to "Set Load to 100%" still works after the shutdown date. So users can still revive a dead core if their analysis wants it so.

stdye:
or just merge the sucker and some PIA fledgling hack can make the minor touch-up changes

stdye:
This looks very good to me. Unless you plan to make more changes, merge at will

DocOtak:
Your "PIA fledgling hack" got a laugh out of me :smile:

PR #254, opened 2021-08-13
separate crust and mantle 

stdye:
guessing it should fail... need to include changes to other files
build did not fail but change crashes site
figured out how to include changes to more than one file in a PR
now try to prevent site crash- success!
now try add percent of total- success!
now try add mantle U- success!
now try add crust Th and mantle Th- success!
now try get percentages correct (IBD only at this point)- success!
now get rid of stale Th code- success! Bonus- the isotope proportions look okay, well built code
now try to improve formatting to prevent overflow data rows- success!
now try to add 235U, now only 238U + 232Th- success!
now try add 40K_beta- success!

This appears to be working. Upgrades could include adding separate crust and mantle downloads and curves on main plot

### DIFF
--- a/src/mantle/index.ts
+++ b/src/mantle/index.ts
@@ -150,6 +150,39 @@ export function mantleGeoSpectrum(
 
     return spec * TargetYears * survivalProbability + ESMUTauContirbution;
   });
+  
+    const geo_mantleTh232 = antineutrinoSpectrum232Th.map((v, i) => {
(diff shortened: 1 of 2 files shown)


stdye:
Please review. If my changes did not damage the code and there are no apparent problems, then you can merge

stdye:
merge when you are good with the changes...

PR #255, opened 2021-08-16
see if <Node inline> works inside <Form>

stdye:
replace verbal description of infinite exposure condition with equation- more compact, now takes one line rather than two


DocOtak:
There was a syntax error in your first attempt maybe try the following:

```javascript
<Node inline>{String.raw`N_{\sigma} \delta B > S`}</Node>
```

PR #256, opened 2021-08-16
prepend symbol in nuisance background rate set

stdye:


stdye:
This works exactly as we want...almost...

If you switch cross section from IBD to ES w/o changing E_min from its virgin value of 1.806 MeV, then E_min sets to 0.

One oddity: When using IBD and changing E_min for the first time, it always goes from 1.806 MeV to 1.8 MeV on the first up or down change.

Made several format changes and added the prepended symbol "b_nusiance" to the value set control

Check it out. Merge it if it looks okay

DocOtak:
That's probably the best way to handle changing the E_min thing. It's very tricky to know "who" changed the value, the set* methods don't care where they are being set, just what value is. 

The "oddity" you've noticed appears to be a Safari specific thing.

stdye:
Da bugga not showin' up. Maybe got stuck somewhere?

Issue #98, opened 2020-10-23
Reactors tab- Core List card- Control all Cores- Turn off all cores


stdye:
Turn off all cores disables the contributions from existing custom cores and prevents adding the contributions of custom cores to the spec stats and spec plot

I am unable to display the spectrum and rate of custom cores without IAEA cores

DocOtak:
Did some digging into this one....

* "Turn off all cores" is doing what I expect, setting the LF of all the cores in the internal core list to 0. This includes any custom cores added before the "Turn off all cores" button is pressed.
* Turning off all the cores does not _remove_ any cores from the list so you are able to find and turn back on individual cores, including custom ones.
* The "add custom" core button will not overwrite existing cores of the same name, so you cannot e.g. add a second "ATR" of the exact same name. This is the an area I think we can improve this by letting you know that some name already exists on the list.

Also keep in mind that the core list by default sorts by distance from the detector, the two research reactor presets are very far away (6000km+) from the default detector site of Boulby.

stdye:
Order of operations seems critical. I got stuck trying to display only the spectrum and rate for a custom core because I added the custom core before hitting the "Turn off all cores" button.

1) The Core Direction Plot with 0 (1) core on displays cos(theta) values from 0 to 6 (2) but cos(theta)>1 is not possible. The cos(theta) range on the plot needs to be fixed to only display values [-1,1].

2) Listing "Use IAEA LF data for all cores" in the drop down menu for "Control all Cores" led me to think that the other two options "Use 100% Load for all cores" and "Turn off all cores" applied only to the IAEA cores. I now see it is coded differently. One way to improve access to "Custom Cores" would be to more fully isolate them from the IAEA cores. How about we make the green "Control all Cores" button work only for IAEA cores? So the green button controls "IAEA Cores" and the blue button controls "Custom Cores."

stdye:
There was a typo in the green "Download Core Spectrum" button. If we separate the controls for IAEA and Custom cores, then download IAEA core spectrum is a green button and download Custom core spectrum is a blue button.

Issue #230, opened 2021-07-30
Filenames for Plotly plot downloads


stdye:
The "Download plot as a png" camera icon button on the upper right of Plotly plots allows users to download the geoneutrinos.org plots shown on the web application.

As previously noted, the filenames, which are now given as "Unknown-n.png" where n is an integer number, are totally generic.

When there are no more substantive issues to address, more descriptive filenames could be implemented.

PR #257, opened 2021-08-19
fix x-axis value ranges, add Plotly filenames- reactors plots

stdye:
The Core Direction Plot with 0 (1) core on displays cos(theta) values from 0 to 6 (2) but cos(theta)>1 is not possible. The cos(theta) range on the plot needs to be fixed to only display values [-1,1].

Above taken from comment on Issue #98

If this works, then will do the same for main spectrum plot

### DIFF
--- a/src/ui/reactors-plots.jsx
+++ b/src/ui/reactors-plots.jsx
@@ -223,7 +223,7 @@ export const CoreDirectionSignalPlots = ({ cores }) => {
   };
   var config = {
     toImageButtonOptions: {
-      filename: 'Core-Directions-signal-vs-cos&theta;'
+      filename: 'Core-Directions-signal-vs-cos(theta)'
     }
   };
   return (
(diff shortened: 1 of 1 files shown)


stdye:
This looks like it works.

Try "Turn off alll cores" and look at "Core Directions w.r.t..." plot on Reactors tab- cos(theta) on x-axis now always -1 to 1

Please merge if OK

stdye:
Aha, there is a small problem. The symbol (dot) for the core at cos(theta) = 1 gets cut in half. Maybe set max vale to 1.05?

stdye:
looks good now on the screen and on the downloaded .png

merge 'em

stdye:
Hmm... Filename changed from 'Unknown.png' to 'newplot-1.png' but not to 'Core-Directions'

Is "config" getting passed?

DocOtak:
Config isn't being passed into this one... add a `config={config}` to the `<Plot>`

stdye:
If they look OK, this PR plus #259, #260, #261 can be merged to retire issues #98 and #230

The main spectrum plot does config directly not as a variable- naming the Plot download filename will done in a separate PR

PR #258, opened 2021-08-20
try fix x-axis values on spectrum plot to be 0-10

stdye:
turning off all cores and geo-neutrinos makes the x-axis values show as 0 to 6 MeV. This should fix the values to always be 0 to 10 MeV.


stdye:
This looks OK to me. Your call on the merge...

DocOtak:
I wonder why it defaults to 6...

PR #263, opened 2021-08-21
Add LF info under titles of Spectrum and Core Directions plots

stdye:


src/ui/plot.jsx
```diff
@@ -5,9 +5,11 @@ import { sum } from "lodash";
 
 import {PhysicsContext} from '../state'
 
+import { Num } from '.'
+
 const evBins = new Float64Array(1000).map((v, i) => i * 0.01 + 0.005);
 
-export function NuSpectrumPlot({ cores, spectrum, detector}) {
+export function NuSpectrumPlot({ cores, spectrum, detector, reactorLF}) {
```
DocOtak, line 12: A new input was added to this function, take a look at how it is "called":
https://github.com/geoneutrinos/reactors/blob/1a21a993571349d9cb7fe604a9a03b56ebb61bf2/src/App.js#L190-L194

stdye: ${} inserts variable to string

but variable not found so site crashes... helpful hint directs problem to /src/App.js

need to add "reactorLF={reactorLF}" between lines 193 and 194

so this commit will crash site until App.js is modified

not sure if changes must be done in tandem or if serial (App.js then this one) is OK

stdye:
This appears to work. Hopefully, it is not too taxing on render time. Your call on whether or not to merge.

Because the Spectrum and Core Directions plots depend on the selected LF range, this information should appear on the plots. A future upgrade could try to handle the situation where a user overrides the LF for one or more cores. This 'custom LF' condition is not reflected in the date range added by the PR or the existing date range on the spectrum stats card.

PR #264, opened 2021-08-21
CoreDirectionsPlot- add detector to title

stdye:
Currently, the reference direction in the title is ambiguous b/c it only mentions the reactor core but not the detector. A direction is defined by two points not one.

this will cause site to crash when selecting Reactors tab until I add "detector" to the import list in App.js

Hmm.. How to include revised App.js without creating a PR?


stdye:
This appears to work. Your call on merging...

PR #265, opened 2021-08-22
increase rounding threshold from 100 km to 200 km

stdye:
There are several new trade studies looking at the possibilities of detecting cores > 100 km from Boulby (i.e. Heysham, Torness) but < 200 km

### DIFF
--- a/src/reactor-cores/index.ts
+++ b/src/reactor-cores/index.ts
@@ -312,7 +312,7 @@ export function ReactorCore({
     const power = this.power;
     const distsq = dist ** 2;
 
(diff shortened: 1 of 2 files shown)


stdye:
Follow cursor on map actually seems better (faster- less janky) in this PR compared with the current website.

Your call on merging this one...

DocOtak:
This is an optimization that trades memory usage for less CPU time, it's pretty arbitrary. Since the scope of potential target cores is now further away, there probably isn't need to artificially introduce extra uncertainty.

Issue #106, opened 2020-11-06
Add Core List "Sort By: Angle" option


stdye:
Add the option to sort the big list of reactor cores by angle, actually cos(theta), but "Sort By: Angle" is the title.

Since cos(theta) is already calculated to make the Core Direction Plot, I am hoping this addition can be done rather easily.

DocOtak:
Clarification: the desire for "sort by angle" is to behave the same way as the plot above? Where it is is wrt between the detector and the "brightest" core.

stdye:
Let's not waste time on this. It was a passing fancy last November. I started working on it last week more as a way to learn about how the code works than to clear out the Issues list. Because all cores are at or near the surface of the planet, angle scales with distance anyway.

But, yes, the idea was to sort on the cos(theta) values.

We have several very important issues associated with going users better access to the elastic scattering signal that need to be addressed.

PR #269, opened 2021-08-24
add_core_sort_by_angle

stdye:
this is issue #106

see if reactors-corelist.jsx knows what core.x is


stdye:
it knows but reactors-plots.jsx must redefine it with what looks like these lines-

  let coreData = []
  if (first !== undefined){
   coreData = [first, ...rest].map((core) => {
    return { y: core, x: first.cos(core) };
   });

PR #272, opened 2021-08-25
prevent plot from showing out of bounds

stdye:


stdye:
reset axes on the Plotly toolbar places the plot inbounds... refreshing the site returns the right side of the plot outside the card boundary again... it is not evident why this particular plot displays this annoying condition

stdye:
none of this flailing had the desired effect...kicking the problem upstairs

PR #274, opened 2021-08-25
replace E_on with E_HM

stdye:
with the new sigmoid function, the role of E_HM is to set the energy of the efficiency half-maximum. E_on is now misleading


stdye:
too many changes... start over...

PR #279, opened 2021-09-02
add Yongbyon to suggestions

stdye:
Yongbyon Nuclear Science and Weapons Research Center in North Korea with one 5 MWe Magnox (GCR) core


DocOtak:
Awesome, these are easy changes, don't wait for my approval for stuff like this.

PR #283, opened 2021-09-04
Create direction-plots.jsx and plot-Te-spectra.jsx

stdye:
This file will contain-

CoreDirectionPlot, which is currently in plot.jsx

CoreDirectionSignalPlots, which is currently in reactor-plots.jsx

as well as future plots showing information on the direction of neutrino sources from a given detector location


stdye:
How's this, boss?

stdye:
Hmmm... plot not showing on Detector tab. It needs to be visible to see if the upcoming change of the scale from Ev to Te=Ev-1.805 is working.

stdye:
The new T_e spectrum plot on the Output tab is valid for IBD. 

It is currently NOT VALID for ES because scattered electron kinetic energy spectrum is not yet calculated and available for plotting

PR #287, opened 2021-09-08
update 40K, 232Th, 235U half lifes

stdye:
Note added reference in comments


stdye:
This looks like it pretty innocuous and the changes are needed for the new half lives table on the geonu tab

Issue #130, opened 2020-12-19
Extend T_min to IBD rate spectra


stdye:
T_min applies to the scattered lepton, the positron in IBD and the electron in ES. Extending T_min to IBD simply cuts out antineutrino energies less than Enu_min = E_thresh + T_min, where E_thresh = [(m_n+m_e)^2 - m_p^2]/2m_p

stdye:
This could be done by adding a column of kinetic energy data to the Reactor download csv file. Nothing else in the file changes. Only add a column with the kinetic energy of the positron T = E_nu - E_thresh, with E_thresh as given above. The kinetic energy bins would have the same width as the antineutrino energy bins. I think only the first bin needs special treatment because T = E_nu - E_thresh < 0. One "fudging it" solution is to just add the contents of the first energy bin to the first bin with T > 0. A quick scan of this new T list can sum all signals in bins with T > T_min.

stdye:
The elastic scattering T_min slider currently affects only the ES rates (as advertised).

It would help users if this slider affected the IBD rates, too. For example, if a detector like WATCHMAN is not sensitive to positrons with T<2 MeV (E_nuebar > 3.8 MeV), then the signal rate from the Hartlepool cores goes down by a fractional amount that depends on the signal spectrum, the background from all other cores goes down by a different fractional amount due to the different spectral shape, and the geo-neutrino background rate disappears.

Retiring this issue improves the web-app's utility to the sponsored project.

stdye:
When we have the kinetic energy of positrons from IBD, we can convert the detection efficiency from a function of neutrino energy E to a function of kinetic energy T. With this change detection efficiency can apply to both IBD and ES.

Issue #252, opened 2021-08-10
Enable output calculator for ES- Part II


stdye:
Make a surface plot of electron antineutrino energy, E, on the x-axis versus scattered electron kinetic energy, T_e, on the y-axis with number of interactions on the z-axis.

For each antineutrino energy in the input spectrum, calculate the ES Differential Cross Section- T_e (as in the plot on the IBD/ES tab) and make a probability density function, pdf (normalize the area under the curve such that the sum of the entries is 1).

Multiply each entry by the number of interactions in the spectrum at the particular antineutrino energy.

Summing the entries from all antineutrino energies at a given electron kinetic energy makes the T_e spectrum for that antineutrino type.

Repeat for the mu-tau antineutrino energy spectrum.

We need to keep track of both the electron and the mu-tau spectra; they have different angular distributions.

Issue #251, opened 2021-08-10
Enable output calculator for ES- Part 1


stdye:
Add column to IBD output files with kinetic energy of positron (T = E_nu - E_thresh). Use the same 10 keV binning. We had this before, I think. Was it called RatPac output? 

Issue #130 is closely related to this issue.

Marc B showed this plot at the Simulations WG meeting today-
[SimBergevin08192021.pdf](https://github.com/geoneutrinos/reactors/files/7018055/SimBergevin08192021.pdf)
The plot on the left is detection efficiency as a function of **_kinetic energy_**, which he converts to detection efficiency as a function of **_antineutrino energy_* in the plot on the right.

Once we get the IBD and ES rate spectrums as a function of kinetic energy up and running, the old form- efficiency(E)- can be retired and replaced by a new form- efficiency(T).

Note A: The detectable signal in the detector comes primarily from the kinetic energy of the scattered lepton (positron in IBD and electron in ES).

Note B: The detection efficiency on the output calculator is easily converted from a function of antineutrino energy (E) to a function of lepton kinetic energy (T). For IBD, E_on becomes T_on = E_on - E_thresh, while epsilon_max and rho are unchanged.  

Note C: Activating the output calculator for ES requires the kinetic energy distribution of the scattered electrons. This will be covered in a subsequent Part to "Enable output calculator for ES"

PR #290, opened 2021-09-09
Unify eV bins and remove assumptions about bin widths

DocOtak:
This is a work in progress PR so things can be tracked.

One thing I did discover is that our NIU calculations all make assumptions that the bins are 10kev wide (there is a _lot_ of 0.01 being multiplied everywhere)

Another oddity is all the cos theta bins for the physics plots have 1001 "bins". The only thing I can think of as for why is that these plots needed to be on closed interval vs the default of a right open interval.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -287,7 +287,7 @@ function App(props) {
                 <h2> REFERENCE </h2>
                 <p> The model presented by this website is fully documented in this paper{" "}
                 <a href="https://arxiv.org/pdf/1510.05633.pdf">arXiv:1510.05633.v3</a>. 
-                Please referee this paper when using the results of this model in your research papers and presentations.</p>
+                Please reference this paper when using the results of this model in your research papers and presentations.</p>
                 <Visible>
                   <OutputDownload
                     spectrum={spectrum}
@@ -298,7 +298,7 @@ function App(props) {
                 </Visible>
                 <CalculatorPanel cores={cores} spectrum={spectrum} />
                 <h2> ACKNOWLEDGMENT </h2>
-                <p> Development of the model and this web application was supported in part by Lawrence Livermore National 
+                <p> Development of the model and this web application is supported in part by Lawrence Livermore National 
                 Security, LLC. </p>
               </Tab>
             </Tabs>
--- a/src/App.js
+++ b/src/App.js
@@ -36,6 +36,7 @@ import {
   FissionFractionPane,
   FissionIsotopeSpectraPlots,
   //Output Tab
+  KESpectrumPlot,
   OutputDownload,
   CalculatorPanel,
   //solar tab
@@ -287,7 +288,13 @@ function App(props) {
                 <h2> REFERENCE </h2>
                 <p> The model presented by this website is fully documented in this paper{" "}
                 <a href="https://arxiv.org/pdf/1510.05633.pdf">arXiv:1510.05633.v3</a>. 
-                Please reference this paper when using the results of this model in your research papers and presentations.</p>
+                Please referee this paper when using the results of this model in your research papers and presentations.</p>
+                <KESpectrumPlot
+              detector={detector}
+              cores={cores}
+              spectrum={spectrum}
+              reactorLF={reactorLF}
+            />
                 <Visible>
                   <OutputDownload
                     spectrum={spectrum}
--- a/src/App.js
+++ b/src/App.js
@@ -36,7 +36,6 @@ import {
   FissionFractionPane,
   FissionIsotopeSpectraPlots,
   //Output Tab
-  KESpectrumPlot,
   OutputDownload,
   CalculatorPanel,
   //solar tab
@@ -289,12 +288,6 @@ function App(props) {
                 <p> The model presented by this website is fully documented in this paper{" "}
                 <a href="https://arxiv.org/pdf/1510.05633.pdf">arXiv:1510.05633.v3</a>. 
                 Please referee this paper when using the results of this model in your research papers and presentations.</p>
-                <KESpectrumPlot
-              detector={detector}
-              cores={cores}
-              spectrum={spectrum}
-              reactorLF={reactorLF}
-            />
                 <Visible>
                   <OutputDownload
                     spectrum={spectrum}
--- a/src/antineutrino-spectrum/index.ts
+++ b/src/antineutrino-spectrum/index.ts
@@ -29,6 +29,7 @@ function resample(
 ): Float32Array {
   const output = new Float32Array(size).fill(0);
   const binWidth = (stop - start) / size;
+  //TODO assumption about bins
   const sliceSize = Math.floor(binWidth * 1000);
 
   return output.map((v, i) => {
@@ -73,6 +74,7 @@ export const antineutrinoSpectrum238U = resample(
 function rateToFluxCalc(spectrum: number[], crossSection: CrossSectionFunc): number {
   const targets = 1e32;
   const rate_to_flux_n = spectrum.reduce((p, v) => p + v, 0);
+  //TODO assumption about bins
   const rate_to_flux_d = spectrum
     .map((v, i) => {
       return v * crossSection(0.0005 + i / 1000);
--- a/src/antineutrino-spectrum/index.ts
+++ b/src/antineutrino-spectrum/index.ts
@@ -31,13 +31,13 @@ function resample(
 ): Float32Array {
   const output = new Float32Array(size).fill(0);
   const binWidth = (stop - start) / size;
-  //TODO assumption about bins
-  const sliceSize = Math.floor(binWidth * size);
+  const inputBinWidth = 1/1000
+  const sliceSize = Math.floor(binWidth/inputBinWidth);
 
   return output.map((v, i) => {
     return antineutrinoSpectrum
       .slice(i * sliceSize, i * sliceSize + sliceSize)
-      .reduce((p, c) => p + c * 100, 0);
+      .reduce((p, c) => p + c * binCount/10, 0);
   });
 }
 
--- a/src/antineutrino-spectrum/index.ts
+++ b/src/antineutrino-spectrum/index.ts
@@ -7,6 +7,8 @@ import { SECONDS_PER_YEAR } from "../physics/constants";
 
 import { CrossSectionFunc, XSNames, CrossSection, crossSection } from "../physics/neutrino-cross-section";
 
+import {binCount} from "../physics/bins"
+
 //interface RateToFlux {
 //  IBDSV2003: number;
 //  IBDVB1999: number;
@@ -30,7 +32,7 @@ function resample(
   const output = new Float32Array(size).fill(0);
   const binWidth = (stop - start) / size;
   //TODO assumption about bins
-  const sliceSize = Math.floor(binWidth * 1000);
+  const sliceSize = Math.floor(binWidth * size);
 
   return output.map((v, i) => {
     return antineutrinoSpectrum
@@ -50,25 +52,25 @@ export const antineutrinoSpectrum40K = resample(
   antineutrinoSpectrum40KData,
   0,
   10,
-  1000
+  binCount
 );
 export const antineutrinoSpectrum232Th = resample(
   antineutrinoSpectrum232ThData,
   0,
   10,
-  1000
+  binCount
 );
 export const antineutrinoSpectrum235U = resample(
   antineutrinoSpectrum235UData,
   0,
   10,
-  1000
+  binCount
 );
 export const antineutrinoSpectrum238U = resample(
   antineutrinoSpectrum238UData,
   0,
   10,
-  1000
+  binCount
 );
 
 function rateToFluxCalc(spectrum: number[], crossSection: CrossSectionFunc): number {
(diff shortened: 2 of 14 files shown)


stdye:
Our rate (R) in NIU calculations are "numerical integrations" w.r.t. energy of the differential rate (dR/dE) spectra. Ideally, we would compute the definite integrals analytically but the functions are too nasty to do this in closed form. So we are stuck with numerical integration.

Currently, our numerical integration uses a sampling "frequency" of 1/(10 keV). Stated inversely, we use a "periodicity" of 10 keV.

Our numerical integration approximates the area under the dR/dE vs E spectra curves by summing the areas of rectangles. Each rectangle has a width (deltaE = E_r - E_l, where the subscripts refer to the left and right edges, equivalent to the period) of 10 keV and a height of dR/dE evaluated halfway between the left and right edges of the rectangle. Using the value of dR/dE_l at the left edge (lower sums) underestimates the area and using the value of dR/dE_r at the right edge (upper sums) overestimates the area. We try to improve the estimate by using the value of dR/dE_m halfway between the left and right edges. Using the "trapezoidal rule" better estimates the area but requires evaluating dR/dE at _both_ the left and right edges of the rectangle, requiring twice as many calls to the dR/dE function as we now use.

We should try switching to using the trapezoidal rule- deltaArea = (E_r - E_l) * 1/2 (dR/dE_l + dR/dE_r) and see if the code still executes fast enough.

Our current method of using dR/dE_m is equivalent to the trapezoidal rule _only_ if the function is linear over deltaE.

stdye:
What can I do to help move this along? We need to get the Detector Efficiency function working on the IBD and ES kinetic energy spectra before the autumnal equinox (Fiscal Year 21 is ending and the Quarterly & Final reports are due 9/24).

stdye:
The cos_theta values must include both -1 (cos_theta[0]) and +1 (cos_theta[1000]) so there is no offset and cosTbins = 1001

stdye:
binAlign could be defined in terms of IBD_THRESHOLD

binAlign = IBD_THRESHOLD - Math.floor(IBD_THRESHOLD) - 0.5 * binWidth

stdye:
FYI- We might want to define E_thresh only once.

On lines 65, 87 in physics/neutrino-cross-section.ts it is E_thresh

On lines 28-30 in /physics/derived.ts it is IBD_THRESH

stdye:
Now is a good time to make the positron kinetic energy distribution dR/dTe for the two IBD cross section functions.

This only requires shifting the x-axis values by the constant IBD_THRESH

Te = Ev - IBD_THRESH

The y-axis values remain unchanged

dR/Te = dR/dEv

The distributions are available for the detector efficiency function and plotting on the Output tab.

Once these are up and running, we add the electron kinetic energy distribution dR/dTe for the ES cross section functions

stdye:
BTW- Lines 40 and 82 in src/antineutrino-spectrum/index.ts look like they have remaining assumptions about energy binning

DocOtak:
So we have a function that shifts the array indices by whatever the IBD threshold is. What do I do with that?

stdye:
Plot it on the Output tab (see PR #283- dR/dT vs T plot) and make function available for multiplication by the detector efficiency function.

Detector efficiency function becomes eff(T) rather than eff(E)

DocOtak:
Can you check the latest build?

stdye:
Nice.

The shifted energy binning is only for IBD. This build applies it to the ES and IBD differential rates.

After this quick fix, add the dR/dT vs T spectrums for ES

stdye:
We could get fancy and overlay the detector efficiency function on the new dR/dT vs T plot with separate y-axis labeling on the right side. This could allow us to cut out the Detector Efficiency plot

DocOtak:
Hm... are you able to see where this shift is being applied to ES? Right now the entire detector efficiency function is skipped for ES. Do we need to apply the detector efficiency to both ES and IBD equally after IBD has been shifted to positron energy?

stdye:
I looked on the spectrum plot after switching to an ES cross section. Plotly shows the x,y values on scroll over. The energy bins had the same shift as IBD.

The detector efficiency function applies to the kinetic energy distributions for both IBD and ES.

DocOtak:
You mean the bin center? I wasn't able to easily add the bin as a property of the PhysicsContext without making what was basically a circular dependency. The number of bins and their size/center is the fundamental assumption of this web app, so it's rather difficult to remove that assumption... 😞 but good progress has been made on this PR I think.

It looks like the total output is shifted by 0.6 NIU for ES with the different bins. I'd argue that that our reactor luminosity model for energies lower than the IBD threshold are significantly more uncertain.
---
For the efficiency function, my interpretation is that when in IBD mode, the spectrum should be shifted to be positron kinetic energy, then detection efficiency applied. And in ES mode, it's already kinetic energy (of electrons right?) so the detector efficiency is applied directly to that spectrum.

stdye:
The new KE spectrum plot for IBD on the Output tab also shows some shift but it should not. The energy bins on the KE plots should be just like on Ev spectrum plots currently on the live version of the web app.

On the live site, the first non-zero bin on the IBD Ev plot is 1.815 MeV, and it is 0.005 MeV on the ES Ev plot.

On the KE plots, the first non-zero bin on the IBD and ES plots looks to be 0.001066207 MeV

stdye:
We do not want to change the bin width/center. All we need is to shift the IBD antineutrino energy scale a tad. But only the IBD antineutrino energy scale.

stdye:
For 1000 bins between 0 and 10 MeV, the KE spectra for IBD and ES should have their first value at 0.005 MeV.

The detector efficiency function doesn't need to know if it operates on IBD or ES.

stdye:
Could reactors/bin.ts export two sets of bins? 

An original one for ES antineutrino energy (first bin center at 0.005 MeV) and a shifted one for IBD antineutrino energy (first bin center at 0.001066207 MeV, making the first non-zero entry at bin center E_thresh + 0.005 MeV)?

To make the IBD KE spectrum, we could use the original one with KE = Ev - E_thresh and starting with the first non-zero entry.

The ES KE spectrum is somewhat more involved but does not require any shift of the energy scale. It is described in Issue #291

stdye:
Moving forward there appear to be several options:

1) Retain original binning (10 keV bins starting at 0) and sampling at the center of the bin (5, 15, 25, ... keV) using the midpoint rule for the numerical integration

2) Retain original binning (10 keV bins starting at 0) but switch to sampling at the bin edges (0, 10, 20, ... keV) and switch to using the trapezoidal rule for the numerical integration

3) Keep track of two bin sets each 10 keV wide, one with an energy offset for IBD and one without an energy offset for ES. Continue using the midpoint rule for the numerical integration

4) Decrease the bin width by an integer factor- 10/2, 10/5, or 10/10 keV- and see if processing speed is adversely compromised. Continue using the midpoint rule for the numerical integration. With memoization maybe the reduction in speed is tolerable.

5) Decrease the bin width by an integer factor- 2, 5, or 10- and see if processing speed is adversely compromised.  Switch to edge sampling (0, 10, 20, ... keV) using the trapezoidal rule for the numerical integration. With memoization maybe the reduction in speed is tolerable

What other options are there? Please share your thoughts on these options

stdye:
When deciding which option to go forward with, we should consider external constraints. Essentially, we are concerned with introducing errors during numerical integration. Errors arise from mismatched sampling and discontinuities in the sampled functions. Here are the external constraints

a. Discontinuity in the IBD cross section function at E_thresh = 1.806066207 MeV

b. Geo-neutrino spectra are given by Enomoto at the midpoints of 1 keV bins- 0.5, 1.5, 2.5, ... keV- and due to the beta decay endpoints there are discontinuities in the spectra

All other input values come from continuous (smoothly varying) functions of energy that we can sample at any energy we please. These include the fission spectra, oscillations, ES cross sections, IBD cross sections for Ev > E_thresh

stdye:
forgot to include 8B solar nu spectrum constraint- listed here and added to previous comment

c. 8B solar neutrino spectrum is given by Winter and Friedman at 0, 100, 200, 300, ... keV up to 15,500 keV

stdye:
[EnergyBin_webapp.pdf](https://github.com/geoneutrinos/reactors/files/7157912/EnergyBin_webapp.pdf)

The discontinuity in the IBD cross section function at E_thresh at 1.806066207 MeV is nearer to the 1.805 MeV midpoint with the original binning than it is to the edges at 1.80 and 1.81 MeV. Therefore, the approximation of the rate from the threshold bin using the trapezoidal rule should be better than using the midpoint rule

DocOtak:
I played around with a bunch of settings today to see how it changes things... some findings:

* The bins are deep deep in the web app such that the "hot reload" (where I can see live changes after editing the code) would not work, I had to completely refresh the app.
* The above issue affected the quick investigation I did when we were on the zoom call. Shifting the bins so that the IBD threshold is at an edge causes a 0.1% change in total NIU for the default core location (I had to use the downloaded output to figure this out).
* I set the bin count to 10000 (1kev), total signal output is also basically the same. Though at the low energies the difference is around 10% because the cross section gradient is so large at those energies.

Having anything which depends on the bins changing or being swapped depending on IBD or ES mode would take a while to implement.

I'm not sure what to do...

Don't these energies just get eliminated when any reasonable (real world) detector efficiency values get applied?

stdye:
This is helpful. Of the options above-

1. Status quo- original binning with poor handling of IBD threshold due to proximity to bin midpoint
2. Original binning with trapezoidal rule improves handling of IBD threshold
3. Impractical
4. Increase number of bins and keep using the midpoint rule- What about speed? Increased resolution dilutes IBD threshold problem but midpoint rule does better when a discontinuity like the IBD threshold is near a bin edge
5. Increase number of bins and change to using the trapezoidal rule- What about speed? Increased resolution dilutes IBD threshold problem but trapezoidal rule does better when a discontinuity like the IBD threshold is near the bin center

Making the electron KE distribution for ES is done by sampling the cross section values at the bin edges rather than the midpoint. This is essentially the trapezoidal rule.

I am leaning towards changing to the trapezoidal rule.

The ES kinetic energy distribution has to get done very soon.

stdye:
Let's remove all the offset and bin aligning stuff. We have learned that it is not practical to carry two sets of bins. Let's go back to the original 10 keV bins that start at 0.

Then let's try increasing the bin count to 2000 and check on the performance. This puts a bin edge at 1.805 MeV and the center of that bin at 1.8075 MeV. This puts the IBD threshold energy 1.80606607 MeV closer to a bin edge than to a bin center, which is good for the midpoint rule. If performance degrades too much, it is now easy to go back to 1000 bins.

After this quick test, we need proceed directly to implementing the ES kinetic energy distribution and making the detector efficiency function operate correctly on the IBD and ES kinetic energy distributions.

stdye:
Compared with live version. Reactor rates unchanged. GeoNu rates change a little bit. Follow cursor a bit slower with 2000 bins.

What do you see?

DocOtak:
Yep, lots of jank and not much difference. Could you check the output download files?

stdye:
The files look good with correct energy binning. The bin width looks like it is hardwired in the filename. No big deal.

Let's stick with 1000 bins. 

Make the ES scattered electron kinetic energy distributions ASAP

stdye:
New vs old 1000 bin versions give different GeoNu totals (~2% lower with new). Reactors stay the same

stdye:
From comparing AntiNu and GeoNu output files, the new energy bins look like they are shifted lower by 5 keV (new = old - 5 keV). At least the listed bin center energies are different new vs old. So are the values in the populated bins, suggesting the rates were calculated differently, possibly using different energies.

stdye:
The reactor and GeoNu rates are now in better agreement (new vs old) after changing to midpoint. For some reason it was set to lower edge.

However, the new mantle rate is lower than the old rate but the new crust rate is the same as the old rate. Also the new Th/U and K/U ratios are higher than the old ratios. This suggests the rate-to-flux calculations may have changed.

DocOtak:
The differences in mantle rates was due to not having the updated/changed isotopic half life data in this PR, I've rebased this PR so they are up to date now. The bin alignment being set to the lower edge was my interpretation of "10 keV bins that start at 0".

stdye:
The lower edge integration changed the GeoNu rates more than the reactor rates because there are discontinuities in the GeoNu spectra but not in the reactor spectra. Both the IBD rates are corrupted slightly by the cross section discontinuity.

It is curious that the crust rates were the same while the mantle rates were different due to using different isotope half lives. Crust fluxes need to be rescaled with changes to isotope half lives, if they are not already.

Onward to producing and plotting the electron kinetic energy distributions from ES!

DocOtak:
Is this ready to merge?

stdye:
No! Just showing the carrots when it is ready to merge.

DocOtak:
What is left to do on this?

stdye:
KE distribution of electrons in ES.

The ES distributions on the KE plot are rate vs Ev (same as main plot) but we want rate vs Te. We need to add code to make the new distributions. For a description of what to do see Issue #291

stdye:
Closed (superseded) Issues #251 #252 might provide further clarification

stdye:
and we need to resolve Issue #130 in this PR

stdye:
Can you remove the changes to output-calculator.jsx that are included in this PR? If so, we might be able to merge this good work on the bins

stdye:
Looks good to me. Merge at will...

PR #297, opened 2021-09-15
import half lives, add scale factors, scale fluxes

stdye:
Scales crust fluxes to account for revisions to isotope half lives


stdye:
This looks to be working as intended. If it checks out with you, merge at will and force push to eV bins monster PR

PR #298, opened 2021-09-23
Plot B8 ES rates as KE

DocOtak:
Check the solar tab for a plot of the scattered B8 energies.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -264,12 +264,9 @@ function App(props) {
               </Tab>
               <Tab eventKey="solarnu" title="SolarNu">
                 <Visible>
-                  <Boron8KEPlot boron8={boron8} />
                   <AnalemmaPlot detector={detector} cores={cores} reactorLF={reactorLF}/>
                   <Boron8SpectraPlot boron8={boron8} />
-                  <CrossSectionPlotsNormal />
-                  <DifferentialCrossSectionPlotsNeutrinos />
-                  <AngularDifferentialCrossSectionPlotsNeutrinos />
+                  <Boron8KEPlot boron8={boron8} />
                 </Visible>
               </Tab>
               <Tab eventKey="ibd/es" title="IBD/ES">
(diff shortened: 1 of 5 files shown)


stdye:
Very nice! It looks like the area under the curve integrates to ~450 NIU as it should. Bravo!

stdye:
Curiously, no, the results using the T_min slider do not make sense.

We expect the rate below T_min to be zero.

Nothing jumps out at me after a quick scan of the new code. It might take a discussion to get this right or lots of comments in the code

stdye:
Some of the solar electron neutrinos oscillate to mu/tau neutrinos. The kinetic energy spectrum of mu/tau neutrinos should be harder than the KE spectrum for electron neutrinos (see the ES differential cross section T_e plot).

Can we also display the KE spectrum for solar mu/tau neutrinos? It could be that we just show one spectra or the other on the plot.

Once we get the plots as we want them, then users should have the option of downloading the KE spectra.

stdye:
Please add column in download-output file for the KE spectrum

Then we check everything and...

Let's get this merged today (9/30 end of FY)

DocOtak:
Ok! let me resolve the conflicts and see what it takes to get the download working.

DocOtak:
Please check the solar download output, this should be "ready to go" if that looks ok.

stdye:
This looks good.

I see one issue: The curves on the total cross section plots on the SolarNu and IBD/ES tabs reflect the current T_min setting but the plot download file names do not indicate the T_min value.

Fix options-

1) add T_min value to plot download filenames

2) display T_min value on the plot- probably on or near the legend

stdye:
The 8B Rate and 8B KE plots have the same T_min issue as the Total Cross Section plots.

Implementing both fix options listed on the previous comment would be ideal

DocOtak:
Let's merge this and followup with a Tmin solution for all the plots that have a dependency on it. I'd lean toward including the information on the plot itself rather than file name so it can be seen in things like presentations or other contexts where you cannot see the file name.

PR #299, opened 2021-09-25
Add mu/tau neutrino ES right and left coefficients and neutrino plots to solar tab

stdye:
needed for 8B solar neutrino calculations


stdye:
This is a work in progress- I need to fix the plot and check the results

stdye:
The exiting code is well constructed, making it very easy to add nu_x cross sections and plots. These additions look good to me. If they pass muster with you, merge at will.

PR #301, opened 2021-09-28
remove ESTotal curve from Total Cross Section plot

stdye:
specify that curves are for antineutrinos


stdye:
Please review and merge as you see fit

PR #305, opened 2021-09-29
try fix typos

stdye:


stdye:
simple and minor improvements... merge at will

then it would be helpful if you would finish implementation of neutrino (non-antineutrino) cross sections

PR #308, opened 2021-09-30
Add T_min information to plots which depend on it

DocOtak:
So far I've only added it to the main spectrum plot for refinement of T_min annotation location and filename pattern. Once we are happy with how it behaves, I'll add it to all the other plots which have a T_min dependency.

### DIFF
--- a/src/ui/physics-plots.jsx
+++ b/src/ui/physics-plots.jsx
@@ -822,7 +822,7 @@ export const CrossSectionPlots = () => {
     },
   ]
   var layout = {
-    title: "Total Cross Sections- Antineutrinos",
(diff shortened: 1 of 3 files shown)


DocOtak:
@stdye Switch the interaction type to one of the ES ones and download the main spectrum plot.

stdye:
Remember Issue #130 

How about adding the T_min = value just after the cross section text to the left of the y-axis label?

stdye:
sorry- dyslexic to the RIGHT of the y-axis label

DocOtak:
That works pretty well for the main plot.

Maybe have the Tmin as a sub title for the other plots (like the average load factor text of the main plot)

stdye:
Yes, under the main title where we put the load factor date range is a good place to put the T_min = value in MeV info on the total cross section and solar nu plots

While you are at it, try fix the main plot

DocOtak:
I think I got all of the plots that have T_min dependencies...

stdye:
Looks good

PR #309, opened 2021-10-01
make plot titles match card headers

stdye:


stdye:
Can you put both 8B plots on the same card?

The SolarNu tab needs to expand to include many other solar neutrino interactions- 7Be, hep, pep, etc... and even some IBD reactions- so condensing the material by limiting the number of card would be helpful

DocOtak:
How would you want the combined card to look? can just stack plots, or maybe even cards within cards.

PR #311, opened 2021-10-02
fix typo: antineutrino -> antineutrino

stdye:
might have to find more files with mis-spelling if this crashes the site


stdye:
Nothing seemed to got broken... review and merge if okay

PR #312, opened 2021-10-03
Update nu oscillation params- NuFit v5.0

stdye:
use latest data- July 2020

modify code to readily use NuFit data


stdye:
Oscillation parameters updated to 2020 values with reference changed to NuFit v5.0

PR #316, opened 2021-10-03
add NO and IO to mass ordering labels

stdye:


stdye:
This introduces the abbreviations NO and IO

PR #320, opened 2021-10-04
Update detector-stats.jsx

stdye:


stdye:
Revisions more explicitly present signal percentages and make the display more uniform across signal sources- IAEA Cores, Geo Neutrinos, etc...

PR #322, opened 2021-10-04
Update physics-osc-params.tsx

stdye:


DocOtak:
What are you trying to do with this one?

DocOtak:
Ah, I see. please use the bootstrap text helper classes for this: add `className="text-right"` to all the elements (th and td) that need it: e.g. `<th className="text-right">`

DocOtak:
You can also apply the style to the entire table by adding `className="text-right"` to the `<Table>` element at the top

stdye:
Hey, you snuck in an edit on me. Just noticed that.

PR #324, opened 2021-10-05
Update physics-plots.jsx

stdye:


stdye:
Well done making the CDF function (we will use it on future upgrades)!

These new CDF plots on the SolarNu tab show the similarities between electron neutrinos and mu/tau neutrinos. The similarities support the assumption of the 8B solar neutrino flux being entirely electron neutrinos

PR #326, opened 2021-10-05
Update neutrino-oscillation.jsx

stdye:


DocOtak:
Looks fine, not a huge fan of abbreviations when there is space for it (NO vs IO) but maybe that change was to the name could fit in the subtitle of the plot?

PR #329, opened 2021-11-27
Update output-download.jsx

stdye:


stdye:
This fixes the units in the headers of the output download files and removes the relic "total" column from the Reactors data output. Merge if the changes to the code are okay.

PR #334, opened 2021-11-30
try limit decimal places- custom detector lat/lon

stdye:


stdye:
As suspected, this affects all detector location Lat/Lon values. As a result, detector location preset values with fewer than 5 decimal places get extra 0's padded onto the displayed values. On the positive side, it does limit the custom detector location Lat/Lon values to 5 decimal places.

The fix needs to be made deeper in the code so that only the follow-cursor-on-map Lat/Lon values have their decimal places reduced from 10 to 5.

PR #335, opened 2021-12-01
Update map.jsx

stdye:


stdye:
This PR reduces the number of decimal places in Lat/Lon values from 15 to 6 for custom detector placement. It does not noticeably slow down the follow cursor function.

If the code changes are acceptable, merge as you wish.

Note: Cloudflare builds are now taking 15-20 minutes and they are not as strict as the Netlify builds.

DocOtak:
We can merge this. It is, however, getting somewhat close to fighting against how numbers work in computers.

The slow build cloudflare issue is something on their end https://www.cloudflarestatus.com/incidents/c3nw000ynl86

stdye:
I agree. Although it produces the desired effect of reducing the number of displayed digits, the method is clumsy at best.

Perhaps a better fix would be to set a new boolean variable "preset" (set to 1 if location is from the list of detector presets and to 0 otherwise) that gets passed to detector-location.jsx. The card uses value={internalDetector.lon} and value={internalDetector.lat} for presets but value={internalDetector.lon.toFixed(6)} and value={internalDetector.lat.toFixed(6)} for custom locations. Or something to this effect... e.g. <option value={internalDetector.lon.toFixed(6)} disabled={preset}>

PR #346, opened 2021-12-19
Reactor_spectra_test

stdye:
this is for comparing reactor signal results with the new Kurchatov Institute 235U and 238U fission isotope spectra to those given by the existing deployment

### DIFF
--- a/src/physics/constants.ts
+++ b/src/physics/constants.ts
@@ -31,8 +31,8 @@ export type VFitIsotopes = {
 // Huber 2011 -> DOI: 10.1103/PhysRevC.84.024617
 // Mueller 2011 -> 10.1103/PhysRevC.83.054615
 export const V_FIT_PARAMS: VFitIsotopes = {
-  U235: [4.367, -4.577, 2.1, -5.294e-1, 6.186e-2, -2.777e-3], // Huber 2011 (phys rev c) table 3
-  U238: [4.833e-1, 1.927e-1, -1.283e-1, -6.762e-3, 2.233e-3, -1.536e-4], // Mueller 2011 table 6
+  U235: [4.353, -4.624, 2.121, -5.335e-1, 6.221e-2, -2.785e-3], // fit of data in Kopeikin 2021 (phys rev d) table 1
+  U238: [2.836, -2.777, 1.376, -3.776e-1, 4.557e-2, -2.072e-3], // fit of data in Kopeikin 2021 (phys rev d) table 1
   PU239: [4.757, -5.392, 2.563, -6.596e-1, 7.82e-2, -3.536e-3], // Huber 2011 (phys rev c) table 3
   PU241: [2.99, -2.882, 1.278, -3.343e-1, 3.905e-2, -1.754e-3], // Huber 2011 (phys rev c) table 3
 };


stdye:
Do not merge. This is only a test at this point

DocOtak:
Do we want to modify the app to support switching between emission spectra?

stdye:
We might want to do this at some point. The Kurchatov Institute results are pretty new. Their paper was published 10/25/21.

stdye:
This new paper (see arXiv link below) uses the KI data to analyze reactor spectra for the main isotopes. The analysis uses a linear interpolation between the logarithm of the data points rather than fitting a degree 5 polynomial to the logarithm of the data points. The web app uses the latter method with the polynomial parameters given by Mueller et al. 2011 and Huber 2011.

https://arxiv.org/pdf/2110.06820.pdf

I am now exploring the differences between the results from linear interpolation and polynomial fit methods. 

This PR uses a polynomial fit from Root that is not published (I did it). Try looking closely at the two new plots at the bottom of the Input tab with this PR. They compare the tabulated data with the Root fits. The fits often miss the data points, especially the bump at high Ev in the 238U spectrum. I am beginning to sour on the polynomial fit method.

If we switch to linear interpolation we can include model spectra other than those from the old Huber-Mueller model. That will be done on a different PR.

stdye:
These plots compare the data values for the 235U spectrum with the fit values for both Huber 201 and Kopeikin et al. 2021 235U.

As you can see from the GeoNu spectra, antineutrino spectra from beta decay can have sharp cutoffs at the endpoint energies. The fits may smooth over these real features. I am leaning more towards using the linear interpolation method.

What do you think?

[U235_H11_pctdif.pdf](https://github.com/geoneutrinos/reactors/files/7772208/U235_H11_pctdif.pdf)
[U235_K21_pctdiff.pdf](https://github.com/geoneutrinos/reactors/files/7772209/U235_K21_pctdiff.pdf)

DocOtak:
I'm a pretty big fan of pchip for interpolating "real data" if you'd like (me) to try that, see the [MATLAB docs](https://www.mathworks.com/help/matlab/ref/pchip.html) for what the algorithm does. The [julia docs](https://juliahub.com/ui/Packages/PCHIPInterpolation/ygqDh/0.1.0) also show what this does. There is a [JS implementation](https://www.npmjs.com/package/slatec-pchip) available.

stdye:
Of course, please get involved.I thought you would have an interpolation routine. I wrote my own routine so a comparison of results is in order.

While researching this issue, one finds a reactor isotope flux model (summation method) that runs from 0.05 to 10.05 MeV!

Errors are stated to be ~5% for 235U, 239Pu, and 241Pu, but rise to 10% for 238U. Using this model below IBD threshold is a significant improvement over blindly extending the polynomial fits into terra incognito. Moreover, researchers are using the spectra at energy above 8 MeV, where the other spectra cut off, to improve the flux estimates.

I attach the paper and the supplemental data here-
[Estienne_et_al_PRL_2019.pdf](https://github.com/geoneutrinos/reactors/files/7781873/Estienne_et_al_PRL_2019.pdf)
[extraMaterial.txt](https://github.com/geoneutrinos/reactors/files/7781874/extraMaterial.txt)

DocOtak:
Did some looking around this morning, this looks like a good "off the shelf" interpolator to start with: [d3-interpolate](https://github.com/d3/d3-interpolate).

stdye:
d3-interpolate looks fine for interpolation. What if we also want extrapolation? (e.g. Giunti et al. extrapolate the Huber, Mueller et al., Kopeikin et al. data, which bottom out at 2 MeV, down to the IBD threshold energy of 1.806 MeV)

Others have asked this question-

https://stackoverflow.com/questions/32841048/does-d3-js-offer-a-method-to-extend-my-dataset-to-the-origin-of-my-graph

Interpolation routine is quite simple (here is my Fortran code for the interpolation points at value ebin between the ith and ith minus 1 data points -w/o prettify- sorry. I wonder if quotation marks disable the default formatter)

" slope=(spec_241_huber(i)-spec_2411_huber(i-1))/(ener(i)-ener(i-1)) "
" sln241h(ebin)=temp1h(i-1)+(energy(ebin)-ener(i-1))*slope "

Extrapolation is similar

stdye:
missed an edit on my previous comment... see edited version

DocOtak:
Working on getting the data into the app, progress happening on a branch https://github.com/geoneutrinos/reactors/tree/reactor_fit

stdye:
**Important:** Interpolation of reactor flux (#nus/fission/MeV) data is between the natural logarithm of the tabulated data values.

This is not evident in the Fortran code pasted above

DocOtak:
This was superseded by #361

PR #350, opened 2021-12-19
Update solar-plots.jsx

stdye:


stdye:
Next steps are: 

use mean from "lodash" to calculate the mean solar intensity for the selected reactor LF date range. The solar intensity (1/au^2) is already being used as the z-axis data on the analemma plot

multiply the solar 8B rate by the mean solar intensity to account for the rate variation due to the distance from the Earth to the Sun during the LK date range

PR #353, opened 2021-12-22
update create react app to 5.0.0

DocOtak:
This updates react-scripts to 5.0.

Since it has better typescript support (I think), it found an issue with the types of the Variable and Derived oscillation params. It looks like the dmsq32Normal and dmsq31Normal were swapped from being given/derived in the NuFIT 5.0 update. I wanted to check to make sure that it is doing the right thing before merging.


stdye:
I did swap these to conform with the PDG calculations. See PR #312

stdye:
The oscillation parameter values tabulated on the Input tab do not change with this PR. Given that we are now passing the values used in the calculations directly to the table, this should be good to merge. Thanks for updating react-scripts!

PR #358, opened 2022-01-08
Update corrections.json

stdye:


stdye:
Well, the changes made to "corrections.json" do not appear online with this PR.

I do not see any instructions, like a README file, so it is not evident what else needs to be done to implement these changes.

stdye:
Using separate lines for each changed value did not work... How to make point changes to the reactor database remains a mystery

stdye:
Original implementation of point corrections was made in PR #253 in which many changes are included.

Maybe my changes are not showing online is because the Cloudflare checks are too loose and the production version is blocked?

stdye:
Hopefully, a full rebuild of the reactor database is not required to make point changes

DocOtak:
You basically cannot do anything with the reactor database without rerunning the python scripts.

stdye:
New location, which increases D_closest by just over 1/4 km, lowers the rate by almost 5%. This is not an insignificant change.  

Previously:
R_closest = 7488.0 NIU
D_closest = 13.20 km (PERRY-1)

Now:
R_closest = 7150.2 NIU
D_closest = 13.48 km (PERRY-1)

stdye:
The newly added locations of Perry-1 and Morton look reasonable when zooming in on the Leaflet map.

The location for Morton needs to be refined once we get better information.

PR #360, opened 2022-01-09
Update corrections.json

stdye:


stdye:
Having trouble replicating your change to the db file. Not enough bread crumbs to follow. I did find the following message:

1 addition, 1 deletion not shown because the diff is too large. Please use a local Git client to view these changes.

Whatever you did (run the python scripts?), do it again to correct the location of Davis Besse-1

stdye:
Done. Make it work.

stdye:
Does it make sense to you that these coord changes put Hartlepool A-1 now closer to Boulby than Hartlepool A-2? Zooming in on the map shows A-2 farther N than A-1. Maybe an illusion due to the Mercator projection?

stdye:
Farther S not N

stdye:
Must be because A-2 moved farther W, too

PR #377, opened 2022-03-14
New Direction Tab

stdye:
new Tab eventKey "directionality" may need definition elsewhere in the code but cannot find anything... try commit and see


stdye:
@DocOtak - please review. If this looks okay as it is, then make it a PR and merge 'em.

PR #382, opened 2022-03-15
removing some parens

stdye:


stdye:
This makes small kind changes to text only

PR #385, opened 2022-03-17
Update map.jsx

stdye:


DocOtak:
It seems that the best/correct way of adding a legend is to create a new "control", found a demo here: https://codesandbox.io/s/how-to-add-a-legend-to-the-map-using-react-leaflet-6yqs5?file=/src/Legend.js

stdye:
Here is one option for the legend. The attributes are it's in a removable layer and it's so ugly user's quickly find it and try to get rid of it!

DocOtak:
Will need to poke at this one later...

stdye:
Can you make a best/correct legend for the main map?

PR #389, opened 2022-03-24
Update map.jsx

stdye:


stdye:
Smaller circle markers on full zoom-in help verify the locations of cores and detector sites. With corrections.json core locations (lat/lon) should be within the core radius. Many locations from INFN/Ultralytics are not close. Check out NOVOVORONEZH-5

PR #390, opened 2022-03-24
Add Shutdown core layer

DocOtak:


DocOtak:
hm... this PR didn't work the way I expected...

PR #394, opened 2022-03-24
Update reactors.csv

stdye:


DocOtak:
Is the preview working as expected?

stdye:
If nothing else, this helps evoke credibility... many cores were far off (see Madras for a particularly egregious placement, and how did San Onofre get built in the ocean?

stdye:
You get the honor of merging...

PR #396, opened 2022-03-25
Update reactors.csv

stdye:
test adding a few shut down dates


stdye:
This looks okay to me on Cloudflare. How about what you see on Netlify?

BTW- I added shutdown dates for two cores in Taiwan that PRIS lists as permanently shutdown but were still active on our core list.

Issue #398, opened 2022-03-25
New cores after latest LF data


stdye:
At this time there are 9 new cores, which are on the grid and spewing antineutrinos, that are not yet in the lagging LF database.

There should be an easy way to display these new cores in a new layer on the main map even if we do not yet include their signals. Maybe use pink circles?

Here are the new cores.

New connections to the grid 2022
--
FUQING-6 | (1000 MW(e), PWR, CHINA) on 1 January
KANUPP-3 | (1014 MW(e), PWR, PAKISTAN) on 4 March
OLKILUOTO-3 | (1600 MW(e), PWR, FINLAND) on 12 March

New connections to the grid 2022
[FUQING-6](https://pris.iaea.org/pris/CountryStatistics/ReactorDetails.aspx?current=938)	(1000 MW(e), PWR, CHINA) on 1 January
[KANUPP-3](https://pris.iaea.org/pris/CountryStatistics/ReactorDetails.aspx?current=1068)	(1014 MW(e), PWR, PAKISTAN) on 4 March
[OLKILUOTO-3](https://pris.iaea.org/pris/CountryStatistics/ReactorDetails.aspx?current=860)	(1600 MW(e), PWR, FINLAND) on 12 March

New connections to the grid 2021
--
BARAKAH-2 | (1345 MW(e), PWR, UAE) on 14 September
HONGYANHE-5 | (1061 MW(e), PWR, CHINA) on 25 June
KAKRAPAR-3 | (630 MW(e), PHWR, INDIA) on 10 January
KANUPP-2 | (1014 MW(e), PWR, PAKISTAN) on 18 March
SHIDAO BAY-1 | (200 MW(e), HTGR, CHINA) on 14 December
TIANWAN-6 | (1000 MW(e), PWR, CHINA) on 11 May

New connections to the grid 2021
[BARAKAH-2](https://pris.iaea.org/pris/CountryStatistics/ReactorDetails.aspx?current=1051)	(1345 MW(e), PWR, UAE) on 14 September
[HONGYANHE-5](https://pris.iaea.org/pris/CountryStatistics/ReactorDetails.aspx?current=997)	(1061 MW(e), PWR, CHINA) on 25 June
[KAKRAPAR-3](https://pris.iaea.org/pris/CountryStatistics/ReactorDetails.aspx?current=986)	(630 MW(e), PHWR, INDIA) on 10 January
[...]

PR #399, opened 2022-03-25
Update reactors.csv

stdye:
TEST- 3 cores (BARAKAH-2, HONGYANHE-5, TIANWAN-6) newly on the grid in 2021 are added (there are 3 more cores new in 2021 not yet added).

How do you want these marked so they can be identifiable for display in a new layer on the main map? A unique shutdown date (2222-01)?


stdye:
Your code is robust! It shows these 3 new cores even though they have no LF data (at least on Cloudflare).

They contribute no signal but one supposes this changes when LF data are present.

Test did not fail. Going ahead then with adding the other 6 cores as noted in Issue #398

DocOtak:
So the behavior coded is:
* If in reactors.csv and not in the DBnnnn.csv files, it will default to all 0 LF (it will also 0 any negative LFs encountered but there is no upper bounds check).
* If in a DBnnnn.csv and not in reactors.csv, it will error and not convert.

stdye:
new core type SHIDAO BAY-1 is 500 MWt HTGR (high-temperature gas-cooled reactor).

 It looks like this breaks the build. 

Either add this new type or the type can be changed to GCR in the list. What do you want to do?

Otherwise, this adds all 9 new cores reported by PRIS as new to the grid after 2020.

What are your comments on highlighting these new cores? New circle color? Note in core Pop-up? Do nothing?

DocOtak:
We can either add it to the "databaseToKnown" mapping (src/reactor-cores/index.ts:443) or if we know power/fission fractions, we can add it as a type we know about.

stdye:
Good point about the fission fractions, which will surely be different that those for the GCR cores. The new HTGR core type can still use the GCR gold color circle.

stdye:
Something else is not correct. HTGR still breaks the build

DocOtak:
The mapping is database (e.g. reactors.csv) to known where "known" is one of the POWER_FRACTIONS or FISSION_FRACTIONS keys (lines 32 to 114)

stdye:
This looks acceptable to me and minimizes the amount of new code like making a new map layer. Merging?

PR #400, opened 2022-03-26
Update reactors.csv

stdye:
add cores nearing criticality and grid connection


stdye:
This adds 8 soon-to-be-operational cores of interest... just so they show on the map

PR #404, opened 2022-03-30
Update map.jsx

stdye:


stdye:
Check out these controls on keeping the map view more focussed on the core and detector circles.

Are there any problems with this PR?

stdye:
This should prevent users from drifting off the "active" map into Sunnyland

Please merge if there are no problems with functionality or presentability

PR #405, opened 2022-03-31
Update index.ts

stdye:


stdye:
Refines positions of some preset detector locations, using ellipsoid heights (notes added in comment field)

PR #411, opened 2022-04-04
Create geonu-luminosities.tsx

stdye:


stdye:
This PR adds table of neutrino luminosities for the GeoNu isotopes to the GeoNu tab

PR #412, opened 2022-04-05
Update constants.ts

stdye:


stdye:
This PR adds GeoNu isotope decay energies as well as mantle mass and mantle geophysical response to constants.ts

Then uses decay energies to (Incompletely) derive isotopic decay heating values in derived.ts 

Heating values need correction for average energy escaping with neutrinos (Eq 33 in the arXiv document)

PR #415, opened 2022-04-05
Create geonu-averageenergy

stdye:


stdye:
This PR:
1) fixes typo in code for energy value (initially 0.005 instead of 0.0005) used when calculating average geoneutrino energies for the isotopes
2) adds table of the average neutrino energies under spectrum plot on GeoNu tab.

PR #417, opened 2022-04-06
Update constants.ts

stdye:


stdye:
need to import "relative atomic mass" for parent and daughter isotopes from ./src/elements/elements.json

stdye:
looks like /ui/elements.jsx might have some clues on how to get the relative atomic mass values from src/elements

stdye:
This PR consolidates Geo Isotope data to one table, deletes the files for separate tables, and tidies up presentation mostly on the GeoNu tab.

Review and if okay merge

DocOtak:
Did some cleanup of an unused import (elementary charge).

Also when you want to write an isotope like <sup>238</sup>U all of the things imported from ui/elements handle this for you, so instead of `<sup>238</sup>U` write `{U238}`

Hopefully everything looks the way it should with these smallish changes.

PR #424, opened 2022-04-23
Update leaflet to 1.8

### LINKED ISSUE #332: Safari browser- "control" + "click"
When the cursor moves over a detector location purple dot marker or a reactor location, the "+" changes to a a hand with a finger pointing at the dot. In my Firefox browser, clicking on the hand pops up a box with the detector or the reactor information and "control" plus "click" pops up the "place detector here" + "place custom core here" + "follow cursor" box. In my Safari browser, simply clicking on the hand again pops up a box with the detector or the reactor information BUT "control" plus "click" pops up BOTH the "place detector here" + "place custom core here" + "follow cursor" box and the detector or the reactor information box.

We should make Safari behave like Firefox.

DocOtak:
It looks like this will fix #332.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "ecef-projector": "^1.0.1",
         "file-saver": "^2.0.2",
         "gh-pages": "^3.1.0",
-        "leaflet": "^1.7.1",
+        "leaflet": "^1.8.0",
         "leaflet-contextmenu": "^1.4.0",
         "lodash": "^4.17.20",
         "papaparse": "^5.3.2",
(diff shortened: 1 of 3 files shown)


stdye:
No change in behavior noted in #332.

Is there any reason not to update leaflet?

DocOtak:
On my Safari; Version 15.4 (17613.1.17.1.13)... it did fix the "two finger click" version, but the ctrl+click does indeed still have the double popup. I wonder if we should look at updating the react-leaflet bindings, those are pretty out of date. The API is quite different now though.

I don't think it will hurt anything to keep leaflet itself up to date... (if everything else is working)

stdye:
I have Safari Version 15.3 (17612.4.9.1.8) and will look into updating it.

Please keep the web app supporting software updated as you recommend.

stdye:
Safari updated to Version 15.4 (17613.1.17.1.13) with macOS Monterey 12.3.1 (21E258) running. The behavior noted in your comment is now observed.

PR #431, opened 2022-06-23
D_closest toFixed(3) for short baseline

stdye:


stdye:
This supports use cases with detectors at short baselines, where distance precision less than +/- 10 m is needed.

Simply changes displayed value from toFixed(2) to toFixed(3)

PR #436, opened 2022-06-29
Reorganize constants

stdye:


stdye:
This PR moves the two PREM calculated values from ./physics/constants to ./mantle/geophysics

MANTLE_MASS
MANTLE_GEOPHYSICAL_RESPONSE

PR #437, opened 2022-06-30
Update geophysics.ts

stdye:


stdye:
This PR moves the isotopic decay heating calculation to ../mantle/geophysics from ../antineutrino-spectrum

PR #440, opened 2022-07-01
Fix isotope decay heating

stdye:


stdye:
This PR corrects the previously miscalculated isotopic decay heating values.

PR #442, opened 2022-07-01
add 40K_ec decay energy and nu luminosity

stdye:
this is a work in progress... first try making EC a thing


stdye:
This PR adds 40K electron capture decay data and heating to refine the mantle radiogenic heating estimate

PR #451, opened 2022-07-22
add some underground sites for possible detector locations

stdye:


stdye:
This PR simple adds a few underground sites as possible detector locations. Elevations still need to be added but none of the sites is very close to a reactor

PR #460, opened 2022-09-24
consolidate sliders

stdye:


DocOtak:
The granularity of this PR was perfect, pretty much the only thing that would make this better is a "location" in the title. e.g. something like "consolidate SN neutrino spectrum sliders"

PR #461, opened 2022-09-24
more consolidation

stdye:


stdye:
This PR organizes the new SnNu tab cards, grouping sliders with event tables

PR #466, opened 2022-09-26
note avg and tot E_nu on plots

stdye:


stdye:
These quantities from "supernova-nus.jsx" need to be made available to "supernova-plots.jsx" for listing on the plots that depend on them-

nueAvgEnrg, anuAvgEnrg, nuxAvgEnrg, nueTotEnrg, anuTotEnrg, nuxTotEnrg

Are they already available?

stdye:
This PR adds subtitles with both total and average neutrino energies to the SnNu spectrum plots

PR #473, opened 2022-09-28
add SnNu spectrum shape parameter to state

stdye:


stdye:
This PR is almost ready BUT one obstacle remains in /supernova/index.ts

see lines 205 and 206

// the factor of 6 in the denominator is (shape_param - 1)! for shape_param=4
// TODO: code the factorial described above

shape_param values are integers between 2 and 6

for shape_param = 5 the factor is 4*3*2*1 = 24
for shape_param = 4 the factor is 3*2*1 = 6
for shape_param = 3 the factor is 2*1 = 2

etc...

DocOtak:
The shape params are definitely only integers between 2 and 6 inclusive? (shape_param - 1)! is suspiciously gamma. If the inputs really are finite. I'd probably prefer a rather short lookup table rather than implement a factorial function if we don't need it.

stdye:
WTF! This is 'perfect'

PR #477, opened 2022-10-03
reorder layer control overlays for cores on map

stdye:


stdye:
This PR sets the default order of map layers to first display Shutdown Reactor Cores and then display Active Reactor Cores.

PR #478, opened 2022-10-07
Tune up SnNu page

stdye:
for comparing output to values for SK in Table 5 in Nakazato et al. PTEP (2018)


stdye:
This PR implements SN neutrino spectrum shape parameters individually for each of the neutrino species, adds a w/o oscillation events for electron elastic scattering, revises the ranges and steps for the E_tot and E_avg sliders.

PR #479, opened 2022-10-09
parameters for nu+16O IBD cross sections

stdye:


stdye:
The .csv files have the parameters for fitting the 16O IBD cross sections as prescribed by Nakazato, Suzuki, Sakura 2018.

The .for file is just some example code, which may or may not be helpful

PR #481, opened 2022-10-12
add fit 16O from Nakazato et al 2018

DocOtak:


DocOtak:
I see the spikiness reported at the edge. It seems to be related to the last term in the fit function growing too fast.

stdye:
These plots compare the data in Table 3 with the fits using Eq 4(a), 4(b) and Table 4 params

[nakazato_nue16o.pdf](https://github.com/geoneutrinos/reactors/files/9779615/nakazato_nue16o.pdf)

[nakazato_nuebar16o.pdf](https://github.com/geoneutrinos/reactors/files/9779618/nakazato_nuebar16o.pdf)

stdye:
Only nue+16O added so far but this seems to work and give reasonable answers... While the new code could be easily cloned for adding nuebar+16O, it could probably be more condensed first

stdye:
To Do List:
- replace spiky 16O IBD cross sections on plot with sum of groups, each using its spike-removing energy threshold
- collapse listed 16O IBD group contributions on IBD events table using "details" option as done for isotope contributions for CEvNS events
- cleanup/optimize new code as needed
- what else?

stdye:
This PR produces the intended results. 

Changes to the tabulated IBD events (including 16O) by using the T_min slider seem to be slower than on the live version with the undesired results (which we should take down).

PR #485, opened 2022-10-19
restore IBD threshold param to group calculator

DocOtak:


stdye:
gets the bad 16O group cross section plots offline

PR #489, opened 2022-10-20
rescale 238U mantle flux control

stdye:
trying to find setting that reproduces "old" version of web app

present slider range has max value too high and step value too coarse

### DIFF
--- a/src/ui/geonu.jsx
+++ b/src/ui/geonu.jsx
@@ -49,9 +49,9 @@ export const MantleFlux = ({ geoFluxRatios, setGeoFluxRatios, geo}) => {
           <Form.Control
             value={geoFluxRatios.U238flux}
             type="range"
-            step={10000}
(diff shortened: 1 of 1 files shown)


stdye:
This PR reduces the available range of the un-oscillated 238U mantle flux to 3x10^6 /cm^2/s, corresponding to a bit more than 32 TW of mantle heating with "typical" Th/U and K/U values. An often quoted value for the radiogenic heating from the crust is ~8 TW. The currently "measured" value of the global surface heat flow is ~40 TW with some of that expected to be "secular" heat residual from the formation of the planet and some expected to be coming from the core. The new maximum flux and the finer step setting better facilitate exploration of mantle models.

At some point it could be interesting to add an option for modeling the geo-neutrino rate from a two-layer mantle, preserving spherical symmetry to keep it simple.

PR #494, opened 2022-10-26
refine lat/lon of Perry-1

stdye:


stdye:
This PR does two things: 

1. makes minor changes to the Lat/Lon of Perry-1 so that the coreCircle is better centered on the OpenStreet Map location of the containment building. 
2. increases the maximum zoom-in by 1 increment (from 18 to 19) to allow more accurate placement of a near detector at Perry-2

PR #495, opened 2022-10-26
make core circles smaller at full zoom-in on map

stdye:


stdye:
This PR makes the core circles smaller (radius = 25) at the new full zoom-in value of 19. Smaller core circles more closely match the size of the typical containment building. This matching allows more accurate placement of core locations.

PR #498, opened 2022-10-26
move LEU from power to fission fractions

stdye:


stdye:
This PR replaces the LEU power fractions with fission fractions from a KamLAND paper. This change reduces the event rate at most sites by about 0.25%

PR #499, opened 2022-10-27
line up core locations w/ OpenStreetMap

stdye:


stdye:
Hopefully OpenStreetMap locations are more "correct" than Google Maps

PR #524, opened 2022-11-14
App.js: Move FissionFractionPane to Reactors tab

stdye:


stdye:
The PR should now be ready for adding oscillation parameter controls to the NuOsc tab

PR #526, opened 2022-11-14
physics-osc-params: complete reference NuFIT data

stdye:


stdye:
The PR updates the neutrino oscillation parameters from NuFIT v5.0 (2020) to NuFIT v5.1 (2021).

With the new data the rate at Fairport from Perry-1 increases by 0.2 NIU out of almost 8000 NIU

PR #532, opened 2022-11-16
App.js: add About tab and uncertainty notes to Stats pane

stdye:
move Reference information and Acknowledgment to new tab


DocOtak (approved):
We might consider a "card" for the stuff under the about tab, but it matches things are they were before so it's ok.

PR #533, opened 2022-11-17
Create about.jsx

stdye:
put About tab content on a card


stdye:
The PR puts the content of the new About tab on a react Card. Revisions and additions to the content can now be made without editing App.js

PR #541, opened 2022-12-19
App.js set default to Kamioka

stdye:


stdye:
Currently the initial view of the map does not include Pacific Northwest/Alaska and East Asia. As such, default detector locations such as Kamioka do not show. This PR starts the initial view zoomed out by changing the zoom value from 2 to 1. Note: A fractional zoom value of 1.5 was tried but it rendered as zoom of 1.

DocOtak:
This is likely dealing with some screen size differences/issues. It might be better to use the [fitbounds method](https://leafletjs.com/reference.html#map-fitbounds) so that the parts of the map we want visible on whatever screen size the user has. We could also try to adjust the "zoom snap" setting if fractional zooms are desired.

PR #543, opened 2022-12-20
reactors.csv correct location of Palo Verde 1

stdye:


stdye:
"Refines" the locations of the three Palo Verde cores to better coincide with the locations shown on Leaflet OpenStreetMap. It remains an open question as to the accuracy of reactor cores locations on OpenStreetMap

PR #549, opened 2023-01-11
map.jsx - alert user to 0 elevation

stdye:
Add note to popup that Place Detector Here uses elevation 0 m


stdye:
Add elevation info to detector-stats.jsx for Distance to closest custom core... BUT customClosestElevation appears to not be a thing... oh, bother

stdye:
This PR facilitates studies of reactor antineutrino signals from custom cores

PR #557, opened 2023-02-23
reactors-corelist.jsx try adding shutdown month

stdye:


stdye:
This PR adds the month of the permanent shutdown date on the core list.

The information is available. Why not post it?

Any thoughts/issues with the format?

PR #565, opened 2023-06-13
Migrate from Create React app To Vite

DocOtak:
It seems that create-react-app isn't maintained anymore, and it has disappeared as recommended from the react documentation as well.

This PR updates the build framework to [vite](https://vitejs.dev/), which is the current apparent successor for a simple web app.


stdye:
Any reason not to merge this PR?

DocOtak:
We can merge I think.

PR #588, opened 2024-03-14
reactors.csv- add ROOPPUR cores under construction

stdye:


stdye:
This PR adds to the database many cores that are under construction. Not yet added are cores that are under construction at new nuclear power complexes in China.

PR #593, opened 2024-03-18
Add PREM

stdye:
This starts the addition of the Preliminary Reference Earth Model (PREM)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -96,7 +96,6 @@ import { reactorAntineutrinoModelReducer } from "./physics/reactor-antineutrinos
 import { defaultBoron8 } from "./solar";
 
 import { PhysicsContext } from "./state";
-import "./mantle/PREM";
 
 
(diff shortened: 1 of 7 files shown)


stdye:
This is work in progress towards adding PREM

stdye:
Now that the code for calculating the PREM density (rho) is working, the next task is to add code that calculates: 
1) the volume of each shell [ vol_shell = 4/3 * pi * ( r_out^3 - r_in^3) ]
2) the mass of each shell [ mass_shell = rho_shell * vol_shell ]
3) the masses of each PREM layer (i.e. inner core, outer core, lower mantle, transition zone, LVZ, LID, Crust), where mass_layer = sum(mass_shell)

stdye:
This works. Just need to replace the entered constants MANTLE_MASS and MANTLE_GEOPHYSICAL_RESPONSE with values calculated internally with the now added PREM model

PR #595, opened 2024-03-28
Add AK135-F 1D radial density earth model

stdye:


stdye:
This starts the addition of a 1D radial density model of the earth. AK135-F is an alternative to PREM with enhanced resolution  (e.g. three crust layers instead to two in PREM).

The model is in the new file AK135-F.csv, which has instructions in the header for calculating the density of each layer.

Once the data are processed to json format with the calculated density values included, then the addition of the model can proceed.

PR #596, opened 2024-03-28
Update to React 18

DocOtak:
This is a maintenance update that updates React to version 18, some of the dependencies needed to be replaced or updated as well:
* The mathjax library changed, the change being, to write math, you now need the `<MathJax>` component with a raw string it in containing math, $$ for block and $ for inline math.
* react-leaflet was updated, this needed some minor changes in how the map component works.

There should be _no_ public user facing changes with this update.


DocOtak:
@stdye Please check the preview build for any math I may have missed, and anything that doesn't work as you might expect it to

stdye:
Looks good. Scanned the pages on every tab.

stdye (approved):

PR #597, opened 2024-04-06
Adding 1D radial density model AK135F

stdye:


stdye:
This PR adds an alternative 1D radial density model for comparison with PREM. GeoNu signal results are unchanged and remain rooted in PREM. Comparison of directional signals, especially from the mantle, for the two models can now be developed.

PR #598, opened 2024-04-08
GeoNu CEvNS addition from PR#577

stdye:


stdye:
Follow cursor on map is super slow when on the GeoNu tab

stdye:
Follow cursor slowness seems to have started with merging of PR#593- Add PREM and got slower with merging of PR#597- Adding 1D radial density mode AK135F. The slowness is not observed in PR#577- Add CEvNS interaction to geonus, so something about the Add PREM merge seems to be the culprit

PR #602, opened 2024-04-09
Update geonu.tsx- why is "memo" imported?

stdye:


stdye:
"Memo" is not used. Maybe it should be?

PR #603, opened 2024-04-10
Geo Models- tidy up

stdye:


stdye:
Taking the array mapping out of the plots speeds up follow cursor.

PR #605, opened 2024-04-12
Geonu spectrums mapping in plot- speed test

stdye:


stdye:
Maybe an imagined effect, but I do not see an obvious speed up for "follow cursor on map" on the GeoNu tab with this PR.

PR #607, opened 2024-04-29
Lunar Neutrinos

DocOtak:
"We choose to go to the Moon"

Work in progress branch and PR for lunar neutrinos this will get larger as the project progresses. Things so far:

- [x] ECEF projector that also works on the moon
- [x] Have a celestial body switcher (assume we are going to Mars next?)
- [x] Map tiles (and different projection params)

Preview Site: https://lunar.reactors.pages.dev/


stdye:
When will the lunar neutrinos software addition be in the beta phase? This should be no later than two weeks from now or May 17. Agreed?

DocOtak:
I think that's a very good goal

stdye:
Earth geo-reactor still enables on "moon"

DocOtak:
> Earth geo-reactor still enables on "moon"

Any suggestions as to what we should have on the moon list?

stdye:
Just surface sited custom core for now.
https://www.space.com/nasa-moon-nuclear-reactor-project-first-phase-complete#:~:text=The%20agency%20set%20requirements%20for,diverse%20approaches%20for%20technical%20review.
[1]
NASA is planning to deploy 40 kWe (120 kWh) reactors with LEU fuel.
Just use the PWR fractional fission rates.

	-----------------------------------------From: "Andrew Barna" 
To: "geoneutrinos/reactors"
Cc: "Stephen Dye", "Comment"
Sent: Sunday May 5 2024 1:25:29PM
Subject: Re: [geoneutrinos/reactors] Lunar Neutrinos (PR #607)

	Earth geo-reactor still enables on "moon"  

	Any suggestions as to what we should have on the moon list? 

	—
 Reply to this email directly,  view it on GitHub [2], or  unsubscribe
[3].
 You are receiving this because you commented.Message ID:  

Links:
------
[1]
https://www.space.com/nasa-moon-nuclear-reactor-project-first-phase-complete#:~:text=The%20agency%20set%20requirements%20for,diverse%20approaches%20for%20technical%20review.
[2]
https://github.com/geoneutrinos/reactors/pull/607#issuecomment-2094992299
[3]

DocOtak:
Since we have the mantle U input as a flux, the TNU values don't actually change when you switch to the moon. The heat generated does switch to something that looks somewhat reasonable. There appear to only be two relevant layers in the model since we will get the crust fluxes from the precomputed table.

We need to have some default lunar mantle values get put in when the user switches to the moon.

stdye:
The lunar radial profile calculations from Briaud et al. 2023 that we are using to characterize the mantle and LVZ layers assume that the crust thickness and density comes from the AVERAGE of the four GRAIL gravity models by Wieczorek et al. 2012. Gridded crustal flux files for this average crust using both Lunar Prospector and Selene (Kaguya) U, Th, K element weight fractions are on LuNu.

DocOtak:
Added a bunch of plots of the input data to the lunar tab, the total web app size is now about 10MB (3MB compressed over the wire).

PR #609, opened 2024-05-23
Lunar model data files

stdye:


stdye:
Use these flux files; they are calculated by averaging the four GRAIL gravity crust models; this is consistent with the radial profile model for the mantle, LVZ, and core.

Make a lunar tab for documentation and maps of the crust thickness, density, element weight fractions, and antineutrino fluxes

PR #610, opened 2024-05-24
Add files via upload

stdye:


DocOtak:
If you are able, upload to this PR the "final state" of data we want available, delete the files that aren't wanted or needed. Switching between models shouldn't be too hard. I'm having trouble remembering the abbreviations.

PR #629, opened 2024-07-03
add revised lunar crust flux model

DocOtak:


stdye:
Almost there! 

Mantle controls should not be active with fixed total HPE masses as in the revised lunar model - see Section V Dye and Barna (2024)

Also note that the revised lunar model puts HPE abundances in the LVZ 10x those in the mantle -see Section V in Dye and Barna (2024)

https://arxiv.org/pdf/2406.00882

stdye:
An alternative lunar mantle control would be an earth primitive mantle (bulk silicate moon) selector. The revised model uses McDonough and Sun (1995) with earth primitive mantle Th=79.5 ppb (+/-15%), U=20.3 ppb (+/-20%), and K=240 ppm (+/-20%). With the lunar crust HPE contents determined by the revised crust model and the BSM contents fixed by the earth's primitive mantle model, the lunar mantle+LVZ HPE contents are constrained by mass balance (BSM - crust). The are several other models for earth's primitive mantle (i.e. Palme and O'Neil (2003), Lyubetskaya and Korenaga (2007)). In this scenario the lunar mantle control would be a pull-down menu of earth primitive mantle models with each specified by a half dozen or so parameters.

stdye:
and... each earth primitive mantle model would have its associated crust fluxes, which could be simply scaled from the default fluxes to account for the different Th/U and K/U

PR #631, opened 2024-07-12
add apollo-15,17 sites Update index.ts

stdye:


stdye:
Using the mouse to place a detector off the top of the moon map (lat>90) crashes the site but not so for the earth map.

PR #633, opened 2024-07-23
Updating reactors database for year 2023

stdye:


stdye:
- some UK cores still reporting negative LFs for some months
- LINGAO cores are "LING AO" in IAEA list
- South Korea now report SHIN-KORI-3 and SHIN-KORI-4 as SAEUL-1 and SAEUL-2

PR #637, opened 2024-12-11
Update neutrino-oscillation.ts

stdye:
update oscillation parameters to NuFit 6.0

### DIFF
--- a/src/physics/neutrino-oscillation.ts
+++ b/src/physics/neutrino-oscillation.ts
@@ -77,12 +77,12 @@ export let oscillation: Oscillation = {
 }
 
 const defaultOscillationParams: VariableOscillationParams = {
-  s2t12: 0.307,
-  dmsq21: 7.41e-5,
-  s2t13Normal:  0.02203,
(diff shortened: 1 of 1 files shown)


stdye:
While the average oscillation survival probability central values for normal and inverted mass orderings are calculated internally, the +/- 1-sigma ranges in these values are calculated with my offline Fortran code.

stdye:
Note The updated parameters include IceCube 24 and SK atmospheric data whereas the previous parameters were w/o SK atmospheric data

DocOtak:
I've verified the params and they look correct. Do merge.

We'll probably want to add the ability to have s2t12 be different for the normal and inverted mass ordering, and probably just for good design support for NO/IO of dmsq21 too.

What are we missing to be able to to calculate the values you've derived offline in Fortran? Just the ± uncertainties on the params?

stdye:
Oops, found something else that needs changing...

stdye:
Yes... just the +/- 1-sigma uncertainties

PR #641, opened 2024-12-17
Preparing for user-defined oscillation params

stdye:
Only adding neutrino oscillation variables and values to neutrino_oscillation.ts. Otherwise there are no changes to the code. See if this compiles before proceeding further.

### DIFF
--- a/src/physics/neutrino-oscillation.ts
+++ b/src/physics/neutrino-oscillation.ts
@@ -230,7 +230,6 @@ export const oscillationReducer = (state:Oscillation, action:OscillationParamsAc
   // recalculate average survival probabilities
   let {
     s2t12Normal,
-    s2t12Normal3SigmaMinimum,
     s2t12Inverted,
     s22t12Normal,
     s22t12Normal3SigmaMinimum,
--- a/src/physics/neutrino-oscillation.ts
+++ b/src/physics/neutrino-oscillation.ts
@@ -8,12 +8,11 @@ export enum MassOrdering {
 
 
 interface VariableOscillationParams {
-  s2t12: number
-  // dmsq21: number
-  // s2t12Normal: number
+  // s2t12: number
+  s2t12Normal: number
   // s2t12NormalSigmaPlus: number
   // s2t12NormalSigmaMinus: number
-  // s2t12Inverted: number
+  s2t12Inverted: number
   // s2t12InvertedSigmaPlus: number
   // s2t12InvertedSigmaMinus: number
   dmsq21Normal: number
@@ -40,8 +39,12 @@ interface DerivedOscillationParams {
   dmsq31Inverted: number
   c4t13Normal: number
   c4t13Inverted: number
-  s22t12: number
-  c2t12: number
+// s22t12: number
+  s22t12Normal: number
+  s22t12Inverted: number
+// c2t12: number
+  c2t12Normal: number
+  c2t12Inverted: number
   s22t13Normal: number
   s22t13Inverted: number
   averageSurvivalProbabilityNormal: number
@@ -62,8 +65,9 @@ type OscillationParams = VariableOscillationParams & DerivedOscillationParams;
 export type Oscillation = OscillationParams & OscillationFunctions & OscillationConfig;
 
 export let oscillation: Oscillation = {
-  s2t12: 0,
-  // dmsq21: 0,
+// s2t12: 0,
+  s2t12Normal: 0,
+  s2t12Inverted: 0,
   dmsq21Normal: 0,
   dmsq21Inverted: 0,
   s2t13Normal: 0,
@@ -77,8 +81,12 @@ export let oscillation: Oscillation = {
   c4t13Normal: 0,
   c4t13Inverted: 0,
 
-  s22t12: 0,
-  c2t12: 0,
+// s22t12: 0,
+  s22t12Normal: 0,
+  s22t12Inverted: 0,
+// c2t12: 0,
+  c2t12Normal: 0,
+  c2t12Inverted: 0,
 
   s22t13Normal: 0,
   s22t13Inverted: 0,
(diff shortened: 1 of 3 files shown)


stdye:
Fruitless effort...

stdye:
TS complains about unused stuff. Stuff to be added now commented out.

stdye:
I can't see the problem with this. Have a look, please

DocOtak:
Looks like the type system is catching some place in the supernova code that is looking for oscillation parameters that have been renamed.

I was looking into some state management techniques that could help with the mess of updating calculated values, this might be a good place to implement them.

Would we be ok with constraining the inputs to be only (with inverted mass ordering and ±uncertainties):

* $\sin^2 \theta_{12}$
* $\sin^2 \theta_{23}$
* $\sin^2 \theta_{13}$
* $\Delta m^2_{12}$
* $\Delta m^2_{32}$
* 
Everything else would get calculated.

stdye:
Mahalo... that's the problem.

It's probably a good idea to discuss code architecture constraints before going too much further.

I want to calculate all values in the Average Survival Probability table so that the site is independent of any external code

stdye:
I didn't see it before but now that I try, yes, updating calculated values is a mess

If constraining to just one of the mass orderings is necessary, then we should use "normal", which is somewhat more favored by the current data

You parenthetically suggest inverted mass ordering but it is not clear why

I'd like to hear about the new state management techniques and your plan for implementing user-settable oscillation parameter values

DocOtak:
Right now we are doing all of our own state management using a few different patterns including react Context, useState, and useReducer, for the oscillation parameters, we are using something called the state reducer pattern (useReducer), we have some state (the current params) and functions that take the state and produce entirely new state. This is all the complications in the switch/case functions.

I had been reading about a [native signals proposal](https://github.com/tc39/proposal-signals) and thought it a very useful pattern that could help reduce a bunch of [boilerplate](https://en.wikipedia.org/wiki/Boilerplate_code) and make the code look more reasonable. If I wanted to use the pattern right now, [MobX](https://mobx.js.org/README.html) seems like one of the better libraries out there. Using something like that, we would define the computed values and let the library deal with updating the values when their source changes, it would look something like this (not real code)

```javascript
let params = {
  s2t12: 0.308,
  get c2t12() {
    return 1 - this.s2t12
  }
}
```
Then if I did a normal assignment like
```javascript
params.s2t12 = 0.309
```
The value for c2t12 would be updated automatically and we can avoid calculating it up front and the following would hold:
```javascript
params.c2t12 == 0.691 // true
```
The actual code would have more going on, but that's the idea...

stdye:
Sounds good to me. I look forward to seeing it in the code. In the meanwhile, I'll stop mucking around.

DocOtak:
Here is what I made in a test to see how well MobX might help here, the entire oscillation parameter implementation seems to collapse into the following:

```javascript
class OscilationParameters {
  s2t12: number;
  s2t23: number;
  s2t13: number;
  dmsq21: number;
  dmsq3l: number; // 3l = 31 for NO and 3l = 32 for IO


  // this is the only boilerplate to make mobx and the constructors work
  // Decorators would have been nice, but browsers don't support them yet
  constructor(s2t12:number, s2t23:number, s2t13:number, dmsq21:number, dmsq3l:number){
    makeObservable(this, {
      s2t12: observable,
      s2t23: observable,
      s2t13: observable,
      dmsq21: observable,
      dmsq3l: observable,

      dmsq31: computed, 
      dmsq32: computed, 
      s22t12: computed,
      c2t12: computed,
      c4t13: computed,
      s22t13: computed,

      survivalProbability: computed,
      averageSurvivalProbability: computed
    })
    this.s2t12 = s2t12
    this.s2t23 = s2t23
    this.s2t13 = s2t13
    this.dmsq21 = dmsq21
    this.dmsq3l = dmsq3l
  }

  get s22t12(){
    return 4 * this.s2t12 * this.c2t12;
  }
  
  get s22t13(){
    const {s2t13} = this
    return 4 * s2t13 * (1 - s2t13)
  }

  get c2t12(){
    return 1 - this.s2t12;
  }

  get c4t13(){
    const {s2t13} = this;
    return (1 - s2t13) ** 2;
  }

  get dmsq31(){
    const {dmsq3l, dmsq21} = this; 
    // If dmsq3l is positve, it is dmsq31, else it is dmsq32 (0 is undefined)
    return Math.sign(dmsq3l) > 0 ? dmsq3l : dmsq3l + dmsq21
  }
  
  get dmsq32(){
    const {dmsq3l, dmsq21} = this;
    // If dmsq3l is positve, it is dmsq31, else it is dmsq32 (0 is undefined)
    return Math.sign(dmsq3l) > 0 ? dmsq3l - dmsq21: dmsq3l;
  }

  get survivalProbability(){
    const {dmsq21, dmsq31, dmsq32, c4t13, s22t13, s22t12, c2t12, s2t12} = this;
    return (Ev: number, dist: number): number => {
      const oscarg21 = 1.27 * dmsq21 * dist * 1000;
      const oscarg31 = 1.27 * dmsq31 * dist * 1000; 
      const oscarg32 = 1.27 * dmsq32 * dist * 1000; 

      const supr21 = c4t13 * s22t12 * Math.sin(oscarg21 / Ev) ** 2;
      const supr31 = s22t13 * c2t12 * Math.sin(oscarg31 / Ev) ** 2;
      const supr32 = s22t13 * s2t12 * Math.sin(oscarg32 / Ev) ** 2;

      return 1 - supr21 - supr31 - supr32;
    }
  }

  get averageSurvivalProbability(){
    const {c4t13, s22t12, s2t13} = this;
    return c4t13 * (1 - s22t12 * 0.5) + s2t13 ** 2;
  }
}
```

I can then make normal and inverted params like this:
```javascript
const oscparamsNO = new OscilationParameters(0.308, 0.470, 0.02215, 7.49, 2.513);
const oscparamsIO = new OscilationParameters(0.308, 0.550, 0.02231, 7.49, -2.484);
```

The code follows the dmsq3l convention in nu-fit and checks the sign to determine if we are dealing with inverted or normal ordering to correctly calculate dmsq31 and dmsq32.

Changing the inputs is trivial and all the changes propagate themselves automatically:

```javascript
oscparamsNO.s2t12 = 0.309
oscparamsNO.c2t12 // 0.691
```

stdye:
Looks much better than the present code.

How about the UI for changing the parameters?

PR #645, opened 2025-02-11
Add 238U and 232Th GeoNu spectra of Li and Xin (2024)

### LINKED ISSUE #644: Add 238U and 232Th GeoNu spectra of Li and Xin (2024)
New GeoNu spectra for 238U and 232Th are available from Li and Xin (2024), as described in 

https://arxiv.org/pdf/2412.07711

These new spectra could be added as an alternative to the current default 238U and 232Th spectra from Enomoto, using a GeoNu spectrum switcher.

Once the spectra are added and the switcher is implemented, then it is a straightforward task to revise the text, tables, and plots on the GeoNu tab.

DocOtak:
Right now this just swaps the data so we can see how it compares while a switcher is built.

Closes #644


stdye:
Would it work to make the calculations with both the Sanshiro spectrums (238U, 235U, 232Th, 40K) and the Li and Xin spectrums (238U, 232Th)? Is the increased number of objects an issue?

stdye:
Hmmm... Can add new isotopic decay heating objects in commit c4acb3d but displaying these objects in the isotope data table fails to build in commit 3079c83. Commit 7410ceb confirms this.

stdye:
This PR adds the Li and Xin spectra, which were published on arXiv last year and recently presented at Neutrino Geosciences 2025, and uses them to calculate values for the rate-to-flux yield factors, average neutrino energies, and isotopic decay heating. These calculated values now appear in the tables on the GeoNu tab.

Please review. This may not be taking the PR in the direction you envisioned. 

However, posting the relevant values for both the Sanshiro and the Li and Xin spectra might be preferable to having a spectrum switcher

DocOtak:
Yeah it's not quite the direction I thought it would go, is it "good enough" though? 

Would the goal to be eventually to allow switching between these sources? How much do we want to be providing the comparative tools/views vs letting folks set up various models, getting the data, and throwing it into their own comparisons?

Technical note, I'm surprised it let you do multiple thead elements (though I'm just noticing them), having more than one is not allowed by the HTML spec

stdye:
No, it's not good enough. Something more like the RA-switcher with a drop-down menu of the spectrum models (Sanshiro, Li&Xin, upcoming, etc.) sitting above tabulated values that change with the model, and results appearing in the main Antineutrino Spectrum plot, with downloadable data files, etc. should be the goal.

PR #649, opened 2025-03-02
Make normal and inverted values of s2t12 and dmsq21

stdye:


stdye:
This PR: 
1) implements normal and inverted values for sin^2(theta_12) and delta_m^2_21;  
2) removes the derived values cos^2(theta_12) and sin^2(2*theta_12)

PR #655, opened 2025-09-15
Make operating detector layer

stdye:
Better viewing if there are separate layers for operational and non-operational detection sites. In the code this functionally similar to active and shutdown reactor cores.

### DIFF
--- a/src/ui/map.jsx
+++ b/src/ui/map.jsx
@@ -248,7 +248,7 @@ export function NuMap({
             <CoreCircles cores={cores} zoom={zoom} shutdownCores={false}/>
           </LayerGroup>
         </LayersControl.Overlay>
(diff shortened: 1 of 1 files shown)


DocOtak:
I think to make this work we would need to add an additional field to the detectors object. Perhaps a status field e.g.:
```js
{name: "Apollo 11", lat:0.67416, lon:23.47314, elevation:0, overburden:0, region:"moon", status:"fictional"}
{name: "IceCube", lat:-89.99, lon:-63.453, elevation:2835, overburden:2200, region:"Antarctica", status:"operational"}
{name: "Fairport", lat:41.756, lon: -81.286, elevation:-433, overburden:1570, region:"North America", status:"shutdown"}
```

stdye:
This PR, which adds a status field to the Detector object, is ready to merge.

PR #657, opened 2025-09-17
adds Load Factor Data for 2024

stdye:


stdye:
Looks like it is working. Please check and do the honors on the Merge if okay

PR #660, opened 2025-09-23
force elevation to 0 in follow mode in detector-location.jsx

stdye:


stdye:
Please merge after review

Issue #632, opened 2024-07-12
Using the mouse to place a detector off the top of the moon map (lat>90) crashes the site but not so for the earth map.


DocOtak:
Using the mouse to place a detector off the top of the moon map (lat>90) crashes the site but not so for the earth map.

_Originally posted by @stdye in https://github.com/geoneutrinos/reactors/issues/631#issuecomment-2224256020_

PR #664, opened 2025-10-06
investigate issue #632- system crash Update map.jsx

stdye:


stdye:
This change limits the lat/lon values when setting the Detection Site to a custom location, using the incremental increase/decrease (up/down) arrows on the card

stdye:
This change limits the lat/lon values when setting the location of a Custom Core, using the incremental increase/decrease (up/down) arrows on the card

stdye:
This change prevents the system crash when using "follow cursor" to place the detection site off the moon map at latitudes 90 < λ < -90

PR #669, opened 2025-10-08
add link to lunar paper on arXiv Update App.jsx

stdye:


stdye:
That works, too. Merge when you like

PR #670, opened 2025-10-08
refine location of proposed lunar settlement Update index.ts

stdye:


stdye:
Not good. It is not so easy to find a color that works for both maps because the earth map has a lighter (white and light blue) background, the moon map has a darker (grey) background, and the core circles already use the basic colors.

PR #676, opened 2025-10-21
adding CEvNS target Tellurium Update geonu.tsx

stdye:


stdye:
This PR adds the the nuclear target Tellurium to the CEvNS signal calculators on the GeoNu tab and the SnNu tab

PR #692, opened 2025-12-17
mantle abundances Update index.ts

stdye:
Begin adding the radionuclide abundance corresponding to the UI flux

### DIFF
--- a/src/ui/geonu.tsx
+++ b/src/ui/geonu.tsx
@@ -343,7 +343,7 @@ export const MantleFlux = ({ geoFluxRatios, setGeoFluxRatios, geo, celestialBody
               <tr>
                 <td>Uranium</td>
                 <td>
-                  <Num v={abundance.U238 / ISOTOPIC_NATURAL_ABUNDANCE.U238} p={1} func={(v) => v * 1e11} />
+                  <Num v={abundance.U238 / ISOTOPIC_NATURAL_ABUNDANCE.U238} p={2} func={(v) => v * 1e11} />
                 </td>
                 <td>
                   <Num v={heating.U238 + heating.U235} p={3} func={(v) => v / 1e12} />
@@ -352,7 +352,7 @@ export const MantleFlux = ({ geoFluxRatios, setGeoFluxRatios, geo, celestialBody
               <tr>
                 <td>Thorium</td>
                 <td>
-                  <Num v={abundance.Th232 / ISOTOPIC_NATURAL_ABUNDANCE.TH232} p={1} func={(v) => v * 1e11} />
+                  <Num v={abundance.Th232 / ISOTOPIC_NATURAL_ABUNDANCE.TH232} p={2} func={(v) => v * 1e11} />
                 </td>
                 <td>
                   <Num v={heating.Th232} p={3} func={(v) => v / 1e12} />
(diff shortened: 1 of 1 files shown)


src/mantle/index.ts
```diff
@@ -242,6 +249,8 @@ export function geoSpectrum(
     crossSection
   );
 
+  const mantleAbundanceU238 = U238flux / ISOTOPIC_NEUTRINO_LUMINOSITY.U238 / mantleGeophysicalResponse
+
   const mantleHeatingU238 = (U238flux / ISOTOPIC_NEUTRINO_LUMINOSITY.U238 / mantleGeophysicalResponse) * ISOTOPIC_DECAY_HEATING.U238 * mantleMass
```
DocOtak: I would make this a function of the mantleAbundance rather than repeating the calculation for that:

```javascript
 const mantleHeatingU238 = mantleAbundanceU238 * ISOTOPIC_DECAY_HEATING.U238 * mantleMass
```

src/mantle/index.ts
```diff
@@ -269,8 +278,9 @@ export function geoSpectrum(
 
   const ThMantleFlux = U238flux * ThURatio * ThMantleFluxIsotopicScale;
 
-  const mantleHeatingTh232 = (ThMantleFlux / ISOTOPIC_NEUTRINO_LUMINOSITY.TH232 / mantleGeophysicalResponse) * ISOTOPIC_DECAY_HEATING.TH232 * mantleMass
+  const mantleAbundanceTh232 = ThMantleFlux / ISOTOPIC_NEUTRINO_LUMINOSITY.TH232 / mantleGeophysicalResponse
 
+  const mantleHeatingTh232 = (ThMantleFlux / ISOTOPIC_NEUTRINO_LUMINOSITY.TH232 / mantleGeophysicalResponse) * ISOTOPIC_DECAY_HEATING.TH232 * mantleMass
```
DocOtak: Same here as above:

```javascript
const mantleHeatingTh232 = mantleAbundanceTh232 * ISOTOPIC_DECAY_HEATING.TH232 * mantleMass
```

DocOtak (changes requested):
Good stuff, see some suggested changes.

stdye:
U238, Th232, and K40 all changed so that heating is explicitly a function of the abundance. Is there any ambiguity with the K40 object names (K40, K40ec, K40beta, etc.)? I am not 100% sure that the K40 abundance calculation is indeed for K40 and not just the K40 beta branch. Can you have a quick look?

Otherwise, merge at will.

DocOtak:
I have maybe confused myself with that K40, are we only including heat from beta? Is ec needed as a separate calcualtion for the total K40 contribution? The `ISOTOPIC_DECAY_HEATING.K40beta` has the [branch ratio included](https://github.com/geoneutrinos/reactors/blob/a4f9930c8997254017ab5630473b995bf02a8bf9/src/mantle/geophysics.ts#L47).

stdye:
The K40 abundance might be off by the natural abundance of 238U, which is very close to 1. Both Th/U and K/U are elemental abundance ratios. The natural abundance of 232Th is also almost 1, making small errors in the 232Th abundance hard to spot.

Try setting K/U = 1e4 and adjust the 238U flux so that the abundance of 238U is as close to 10 ppb as you can get it. Then scale this 10 ppb by the K/U ratio and the natural abundance of 40K. Compare this with the abundance of 40Kbeta.

It looks like the value scaled from the 238U abundance (11.7035)  is less than the 40Kbeta (11.79) abundance by a factor very close to the natural abundance of 238U (0.9927).

Maybe you can get this sorted. I gotta hit the rack.

stdye:
Merge at will if all good

PR #697, opened 2025-12-19
Update detector-location.jsx

stdye:
Try something else...


stdye:
This doesn't seem to change anything

PR #699, opened 2025-12-22
Update geonu.tsx

stdye:
begin adding code for signal from layered mantle relative to uniform mantle

### DIFF
--- a/src/ui/geonu.tsx
+++ b/src/ui/geonu.tsx
@@ -591,9 +591,7 @@ export const LayeredMantleFlux = () => {
         <Table>
           <thead>
             <tr>
-              <th></th>
-              <th>Constant Heating</th>
-              <th></th>
+              <th colSpan={3} >Constant Heating</th>
             </tr>
             <tr>
               <th>Mass Fraction</th>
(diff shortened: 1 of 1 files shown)


DocOtak:
I haven't touched these since this draft PR got started (maybe 3 days ago?)

src/ui/geonu.tsx
```diff
@@ -29,8 +29,20 @@ import {
...
 } from "../mantle/geophysics";
 
-const {K40, Th232, U235, U238} = ElementsUI
+import {
+  layerMasses,
+  layerGeoResponse,
+} from "../mantle/PREM";
 
+const massFunc = (radBot, radTop) => {
+  return layerMasses.slice( radBot * 10, radTop * 10 ).reduce((massSum, currentMass)=>massSum + currentMass)
```
DocOtak: The crash is being caused here when the radBot and radTop have the same value, this causes the slice to be empty so reduce throws. This can be solved by setting an initial value (0?) to the reduce function in both here (massFunc) and geoResponseFunc.

```javascript
return layerMasses.slice( radBot * 10, radTop * 10 ).reduce((massSum, currentMass)=>massSum + currentMass, 0) // see the added 0 argument to reduce 
```

stdye:
Perhaps the UI controls should be thickness and enrichment factor rather than thickness and depletion factor. This way the user would control the properties of a single layer rather than two layers. I think this would be more straight forward.

stdye:
This is close to ready.

The Residual Fraction control works well. Users are unable to type in non-numeric characters or numeric characters that put the value out of bounds.

The Layer Thickness control works. Although users are able to type in non-numeric characters, they can be deleted and the calculations recover valid values.

Comments? Suggestions?

stdye:
This PR calculates the surface flux from a fixed amount of a given nuclide distributed in the lower enriched mantle and the upper depleted mantle relative to the surface flux from the uniform mantle distribution. This is the relative flux for iso-heating.

The relative heating for iso-flux is interesting as well. In this inverted calculation, fixing the flux results in an increase in the heating in the layered mantle relative to the uniform mantle.

PR #700, opened 2025-12-31
Update plot.jsx

stdye:
start adding that and phi information to hover over banner on core direction polar plot


stdye:
Merge at will. Note that an initial change was merged inadvertently

PR #729, opened 2026-02-18
reconstructing PR#704 part 1

stdye:


stdye:
**Small discrepancy in reactor signal**
Default settings: Total reactor signal on Detector Stats panel for Jiangmen is 1328.31 NIU, however, the total reactor signal from the core download file is 1329.13 NIU.

stdye:
The small discrepancy is due to PR #731

